PR #53, opened 2018-04-27
Add context menu items

### LINKED ISSUE #38: Contesxt menu item for converting text
Use case: https://github.com/rugk/offline-qr-code/issues/37

How it should behave:
1. User selects text.
2. User right clicks.
3. User selects "Create QR code from text".
4. Popup is opened and QR code is shown.

What should not happen:
* When no text is selected, no context menu entry should be shown. Even an entry "Generate from URL" would be unnecessary as there is the button for that already, which is far easier to be used (as it is only one click away in contrast to a context menu item, which is two clicks away.).

### LINKED ISSUE #48: Right-click menu on icon to open options
Maybe it is possible to add a context menu item on the icon, so when you right-click on it, you can open the settings of the add-on.

ENT8R:
This PR adds two context menu items for two actions:
1. convert the selected text to a QR code by right clicking the selection (fixes #38)
2. right-click menu on icon to open options (fixes #48)

I did some research but I could not find a solution for getting the selected text from a page which begins with `about:`. Somehow Firefox restricts the access to these pages and one cannot get the selected text 😞
I added a TODO comment here:
https://github.com/ENT8R/offline-qr-code/blob/context-menu/src/popup/qrcode.js#L725-L729


rugk:
Wait… no… did not you saw I assigned these issues to me?

I already coded that stuff yesterday, see branch https://github.com/rugk/offline-qr-code/tree/contextmenu.

But I'll have a look at your code. Maybe we can merge them…
Next time, maybe also comment on the issue first.

ENT8R:
Oh... 😄 didn't saw this... I will take a look at your code.

rugk:
So I looked at the code. (and of course compared it with my implementation)

First I like how you did the background.js stuff. It is really split into modules/functions and better structured than my version.

However, the actual sending to the popup and handling there is a thing you did overly complicated. Seems you inject content scripts or so, just for getting the selection. Actually Firefox provides the selection when the context menu entry is clicked.

> I did some research but I could not find a solution for getting the selected text from a page which begins with about:. 
Yeah, of course this is blocked. Same as `addons.mozilla.org`. For security reasons, of course. – So you cannot inject content scripts there.
However, I just tested my method of getting the stuff directly from the API and it works on these sites too…

rugk:
So I'll now take some code from you and merge it into my branch… Let me see… :smiley:

rugk:
BTW the `MessageListener` thing looks interesting. Because we do not need it, as no context script is necessary, I'll won't include it, but maybe once we need it, we could come back to that code…

Also noticed another mayor difference:
* You save the "use context menu" setting in the background JS and query it when opening the popup.
* I just let the background script send a message to change the text be set if it is received. If not,  the default behavior will be triggered.

So I made my version a bit more asynchronous – having the theoretical risk that the wrong QR code is generated first.

rugk:
BTW the `MessageListener` thing looks interesting. Because we do not need it, as no context script is necessary, I'll won't include it, but maybe once we need it, we could come back to that code…

Also noticed another mayor difference:
* You save the "use context menu" setting in the background JS and query it when opening the popup.
* I just let the background script send a message to change the text be set if it is received. If not,  the default behavior will be triggered.

So I made my version a bit more asynchronous – having the theoretical risk that the wrong QR code is generated first. As that never happened in my tests I have not even bothered to implement some change for that… :laughing: 
But I'll do that now.

ENT8R:
> However, the actual sending to the popup and handling there is a thing you did overly complicated. Seems you inject content scripts or so, just for getting the selection. Actually Firefox provides the selection when the context menu entry is clicked.

Well, this makes things so much easier... 😄

> I just let the background script send a message to change the text be set if it is received. If not, the default behavior will be triggered.

That's also a pretty cool solution! I really did it overly complicated... 😏

I just tested it now and it works like a charm 🎉 Thank you 👍

PR #58, opened 2018-04-27
Add a context menu entry if a link is selected

### LINKED ISSUE #55: Context menu entry for link
User:
1. Sees link.
2. Right-click on link.
3. Choose "Generate QR code from link".
-> A QR code should be generated from the link, without the need to actually open/visit the site.

This depends on/extends https://github.com/rugk/offline-qr-code/issues/38.

ENT8R:
This PR fixes #55
screenshot 1

### DIFF
--- a/src/_locales/de/messages.json
+++ b/src/_locales/de/messages.json
@@ -5,7 +5,7 @@
     "description": "Name of the extension."
   },
   "extensionDescription": {
-    "message": "Elaubt, einen QR-Code für jede Webseite oder Tab zu generieren, die du offen hast.",
+    "message": "Erlaubt, einen QR-Code für jede Webseite oder Tab zu generieren, die du offen hast.",
     "description": "Description of the extension."
   },
   "browserActionButtonTitle": {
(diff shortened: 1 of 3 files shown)


ENT8R:
> if I actually both select something and right-click on a link? So e.g. select a part of an URL and right-click on it

Then both actions are shown. Should this be handled differently?

ENT8R:
It looks then like this:
screenshot 2

rugk:
> Then both actions are shown. Should this be handled differently?

Ah, no, now that I am seeing that, I think this is the best thing, it can do. :smile:

src/_locales/de/messages.json
```diff
@@ -68,11 +68,15 @@
   // context menu
   "contextMenuItemConvertSelection": {
     "message": "QR-Code aus Auswahl",
-    "description": "The context menu entry shown for generating QR codes from a slection."
+    "description": "The context menu entry shown for generating QR codes from a selection."
+  },
+  "contextMenuItemConvertLinkSelection": {
+    "message": "QR-Code aus Link erstellen",
```
rugk: As with the English translation, don't include the word "erstellen" here. It is obvious this add-on does not destroy QR codes.

rugk:
Thanks, it all works great.

PR #60, opened 2018-04-27
Add description for the error correction option

### LINKED ISSUE #13: Info button/helper for Error Correction level option
Next to the error correction level there should probably be a [link to Wikipedia](https://en.wikipedia.org/wiki/QR_code#Error_correction), which explains what that is, as it is not obvious for the general user.

Maybe an icon, or **even better**, a [helper](https://design.firefox.com/photon/patterns/helpers.html) as they are called in Firefox' design. A small text descripting it in one sentence and link to Wikipedia would be a good thing.

ENT8R:
This PR fixes #13
unbenannt


src/options.html
```diff
@@ -67,6 +67,9 @@
 						<option data-i18n="__MSG_optionEcQuartile__" value="Q">Quartile (25%)</option>
 						<option data-i18n="__MSG_optionEcHigh__" value="H">High (30%)</option>
 					</select>
+					<span data-i18n="__MSG_optionErrorCorrectionDescr__" class="helper-text">If a QR code is damaged or some parts are missing, it can often still be read by machines.<br>
+						The higher the value the more the machine is able to understand. For more information see <a href="https://en.wikipedia.org/wiki/QR_code#Error_correction">this Wikipedia article</a>
```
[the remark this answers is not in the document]
ENT8R: Yes, I just tested it and it does...

rugk:
As for the text: Maybe the "by maschines" is a bit superfluous, because one thing is clear: We as humans can never read a QR code – no matter what error correction level. :wink: 
And to make it less technical maybe just say "can be scanned" or so(?).

src/common/common.js
```diff
@@ -198,7 +198,7 @@ const Localizer = (function () {
             const translatedMessage = browser.i18n.getMessage(getMessageTag(tag));
             // only set message if it could be retrieved, i.e. do not override HTML fallback
             if (translatedMessage !== "") {
-                elem.textContent = translatedMessage;
+                elem.innerHTML = translatedMessage;
```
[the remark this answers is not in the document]
ENT8R: How do you imagine could this be done with plain text?

src/common/common.js
```diff
@@ -196,9 +196,15 @@ const Localizer = (function () {
         // localize main content
         if (tag !== "") {
             const translatedMessage = browser.i18n.getMessage(getMessageTag(tag));
+            const isHTML = translatedMessage.startsWith("!HTML!");
             // only set message if it could be retrieved, i.e. do not override HTML fallback
             if (translatedMessage !== "") {
-                elem.innerHTML = translatedMessage;
+                if (isHTML) {
+                    const normalizedMessage = translatedMessage.replace(/!HTML!/, "").trim();
```
rugk: Hmm, I am not sure whether trimming the whole string is always want you want. I mean, in the current code base there is no real need for deliberate spaces at the end or start of a string AFAIK, but maybe there will be in the future.

Although not standardized maybe [trimLeft](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/trimLeft) is all right here, because it seems to have widespread support anyway. (Or could it trim only one space after the string, if needed…?)

src/common/common.js
```diff
@@ -209,9 +215,10 @@ const Localizer = (function () {
             if (elem.hasAttribute(currentLocaleAttribute)) {
                 const attributeTag = elem.getAttribute(currentLocaleAttribute);
                 const translatedMessage = browser.i18n.getMessage(getMessageTag(attributeTag));
+                const isHTML = translatedMessage.startsWith("!HTML!");
                 // only set message if it could be retrieved, i.e. do not override HTML fallback
                 if (translatedMessage !== "") {
-                    elem.setAttribute(currentAttribute, translatedMessage);
+                    elem.setAttribute(currentAttribute, isHTML ? translatedMessage.replace(/!HTML!/, "").trim() : translatedMessage);
```
rugk: Hmm, this assumes placeholders and stuff can never contain HTML… and yeah… it's true. So okay.
However, if someone uses !HTML! there the translation string is either invalid or that is actually wanted, so there is no real need to replace it here too, but I am fine with doing it anyway – it does not hurt.

But also here: leftTrim only, possible.

src/_locales/de/messages.json
```diff
@@ -149,8 +149,13 @@
     "description": "Option of error correction level for QR code. See https://en.wikipedia.org/wiki/QR_code#Error_correction."
   },
   "optionErrorCorrectionDescr": {
-    "message": "Auch wenn ein QR-Code kaputt ist, kann er noch von Maschinen gelesen werden. Dieser Wert zeigt an wie viel von dem Code fehlen kann. Für mehr Information siehe <a href='https://de.wikipedia.org/wiki/QR-Code#Kapazit%C3%A4t_und_Fehlertoleranz'>diesen Wikipedia Eintrag</a>.",
-    "description": "The description of the error correction option."
+    "message": "!HTML! Auch wenn ein QR-Code kaputt ist, kann er immer noch gescannt werden. Dieser Wert zeigt an wie viel von dem Code fehlen kann. <br> Für mehr Information siehe $URL$.",
```
rugk: There is no need to use spaces around the `<br>` tag. Some browsers could maybe just do awkward things, so better leave them out.

Or… wait… do we even need this forced line break? Such things may not always look good/destroy layout.
Firefox also does not seem to use these on their settings page.

LICENSE.md
```diff
@@ -81,6 +81,7 @@ The icons `assets/qr-icon*.svg` (used in `src/icons/icon-small-*.svg`) are made
 * [`popup/img/error-white.svg`](popup/img/error-white.svg) by Mozilla, license: [MPL v2.0](https://www.mozilla.org/en-US/MPL/2.0/)
 * [`popup/img/info-dark.svg`](popup/img/info-dark.svg) by Mozilla, license: [MPL v2.0](https://www.mozilla.org/en-US/MPL/2.0/)
 * [`popup/img/info-light.svg`](popup/img/info-light.svg) by Mozilla, license: [MPL v2.0](https://www.mozilla.org/en-US/MPL/2.0/)
+* [`icons/photon/open-in-new-16.svg`](icons/photon/open-in-new-16.svg) by Mozilla, license: [MPL v2.0](https://www.mozilla.org/en-US/MPL/2.0/)
```
rugk: There is no need for a phaeton dir. All current icons are from Photon anyway…

So either move all of them into this dir – or well… just put the new one is this dir.

src/_locales/en/messages.json
```diff
@@ -150,8 +150,13 @@
     "description": "Option of error correction level for QR code. See https://en.wikipedia.org/wiki/QR_code#Error_correction."
   },
   "optionErrorCorrectionDescr": {
-    "message": "If a QR code is damaged it can still be read by machines. This value indicates how much of the code can be missing. For more information see <a href='https://en.wikipedia.org/wiki/QR_code#Error_correction'>this Wikipedia article</a>.",
-    "description": "The description of the error correction option."
+    "message": "!HTML! If a QR code is damaged it can still be scanned. This value indicates how much of the code can be missing. <br> For more information see $URL$.",
```
rugk: Also this `!HTML!` "trigger string" (or however one wants to call this) needs soem documentation in the Contributing.md.

LICENSE.md
ENT8R: I just didn't put the icon in the `popup/img/` directory because this icon is not really related to the popup but rather to the options/settings. But maybe it will be needed later so I will move it into this folder...

src/common/common.js
ENT8R: Alright. I was not aware of this method.

CONTRIBUTING.md
```diff
@@ -14,6 +14,15 @@ For more details, [see the official docs](https://developer.mozilla.org/Add-ons/
 
 The English "you" should be translated in a personal way, if the target language differentiates between "you" for "anybody"/"they" and "you" for "the user of this extension". In German, that e.g. means you can translate it with "du [kannst etwas machen]" instead of "man [kann etwas machen]".
 
+### Translations which use HTML
+
+If you need to add HTML to a translation, please add `!HTML!` in front of the sentence. It can only be parsed properly if you do this. Otherwise it will display the HTML as normal characters. A possible sentence should look like this:
+```js
```
rugk: It's JSON, not JavaScript :smile:

CONTRIBUTING.md
```diff
@@ -14,6 +14,15 @@ For more details, [see the official docs](https://developer.mozilla.org/Add-ons/
 
 The English "you" should be translated in a personal way, if the target language differentiates between "you" for "anybody"/"they" and "you" for "the user of this extension". In German, that e.g. means you can translate it with "du [kannst etwas machen]" instead of "man [kann etwas machen]".
 
+### Translations which use HTML
+
+If you need to add HTML to a translation, please add `!HTML!` in front of the sentence. It can only be parsed properly if you do this. Otherwise it will display the HTML as normal characters. A possible sentence should look like this:
```
rugk: Maybe better use "prepend" here as an exact term. But… no I'll change that after it is merged. There is some wording I'll want to change anyway.

LICENSE.md
```diff
@@ -77,11 +77,11 @@ The icons `assets/qr-icon*.svg` (used in `src/icons/icon-small-*.svg`) are made
 
 ## Other third-party components
 
-* [`popup/img/check.svg`](popup/img/check.svg) by Mozilla, license: [MPL v2.0](https://www.mozilla.org/en-US/MPL/2.0/)
-* [`popup/img/error-white.svg`](popup/img/error-white.svg) by Mozilla, license: [MPL v2.0](https://www.mozilla.org/en-US/MPL/2.0/)
-* [`popup/img/info-dark.svg`](popup/img/info-dark.svg) by Mozilla, license: [MPL v2.0](https://www.mozilla.org/en-US/MPL/2.0/)
-* [`popup/img/info-light.svg`](popup/img/info-light.svg) by Mozilla, license: [MPL v2.0](https://www.mozilla.org/en-US/MPL/2.0/)
-* [`icons/photon/open-in-new-16.svg`](icons/photon/open-in-new-16.svg) by Mozilla, license: [MPL v2.0](https://www.mozilla.org/en-US/MPL/2.0/)
+* [`common/img/check.svg`](common/img/check.svg) by Mozilla, license: [MPL v2.0](https://www.mozilla.org/en-US/MPL/2.0/)
```
rugk: Oh, yeah, good catch… This stuff is always happening when you move files around. :smile:

CONTRIBUTING.md
ENT8R: Alright. Thank you!

rugk:
So I changed the style a bit, e.g. I saw that the focused style has a rounded border in the image – there is just no description. (https://github.com/FirefoxUX/photon/issues/297)

Also I am not sure whether we need the "external link" image here. I mean, yes, it's external, but it is obvious and all settings sites in Firefox also do not have that.

ENT8R:
Yes, looks also good to me. 

> Also I am not sure whether we need the "external link" image here.

If you want to, you can remove it

PR #64, opened 2018-04-30
Allow messages to be dismissable

### LINKED ISSUE #11: MessageHandler: Allow messages dismissable
It would be great to be able to dismiss a message, e.g. the "success" message shown when the settings are reset to defaults.

Follow the Firefox design: https://design.firefox.com/photon/components/message-bars.html#dismissible

ENT8R:
This PR fixes #11 
unbenannt

### DIFF
--- a/src/common/common.css
+++ b/src/common/common.css
@@ -71,11 +71,21 @@ body {
 .success::before {
   background-image: url('/common/img/check.svg');
 }
+
 .icon-dismiss {
   background-image: url('/common/img/close.svg');
   position: absolute;
-  right: 10px;
+  right: 0px;
   cursor: pointer;
+}
+.icon-dismiss:hover {
+  background-color: rgba(12, 12, 13, 0.1);
+}
+.icon-dismiss:active {
+  background-color: rgba(12, 12, 13, 0.2);
+}
+.icon-dismiss:focus {
+  box-shadow: 0 0 0 1px #0a84ff inset, 0 0 0 1px #0a84ff, 0 0 0 4px rgba(10, 132, 255, 0.3);
 }
 
 .info {
--- a/src/common/common.css
+++ b/src/common/common.css
@@ -74,18 +74,21 @@ body {
 
 .icon-dismiss {
   background-image: url('/common/img/close.svg');
-  position: absolute;
-  right: 0px;
+  margin-left: auto;
+  margin-right: 0;
   cursor: pointer;
 }
 .icon-dismiss:hover {
   background-color: rgba(12, 12, 13, 0.1);
+  border-radius: 2px;
 }
 .icon-dismiss:active {
   background-color: rgba(12, 12, 13, 0.2);
+  border-radius: 2px;
 }
 .icon-dismiss:focus {
   box-shadow: 0 0 0 1px #0a84ff inset, 0 0 0 1px #0a84ff, 0 0 0 4px rgba(10, 132, 255, 0.3);
+  border-radius: 2px;
 }
 
 .info {
--- a/src/common/common.css
+++ b/src/common/common.css
@@ -74,6 +74,19 @@ a:active {
   align-items: center;
 
   z-index: 2;
+
+  /* fade-in transition */
+  /* follow https://design.firefox.com/photon/motion/duration-and-easing.html */
+  opacity: 1;
+  max-height: 100px;
+
+  transition: opacity 150ms cubic-bezier(.07,.95,0,1),
+              max-height 200ms cubic-bezier(.07,.95,0,1);
+}
+.message-box.fade-hide {
+  max-height: 0px;
+  opacity: 0;
+  min-height: 0px;
 }
 
 .error {
@@ -115,6 +128,9 @@ a:active {
 
   margin-left: auto;
   cursor: pointer;
+
+  /* some animation on hover */
+  transition: background-color 150ms cubic-bezier(.07,.95,0,1);
 }
 .icon-dismiss:hover {
   background-color: rgba(12, 12, 13, 0.1);
--- a/src/common/common.css
+++ b/src/common/common.css
@@ -80,8 +80,9 @@ a:active {
   color: #ffffff;
   background-color: #d70022;
 }
-/* icons for the message box */
-.message-box::before, .icon-dismiss {
+
+/* icons for the message boxes */
+.message-box::before {
   display: inline-block;
 
   /* fixed size */
@@ -106,9 +107,13 @@ a:active {
 }
 
 .icon-dismiss {
-  background-image: url('/common/img/close.svg');
+  padding: 4px;
+
+  box-sizing: content-box;
+  width: 16px;
+  height: 16px;
+
   margin-left: auto;
-  margin-right: 0;
   cursor: pointer;
 }
 .icon-dismiss:hover {
--- a/src/common/common.css
+++ b/src/common/common.css
@@ -53,7 +53,8 @@ a:active {
 /* message box */
 /* follows https://design.firefox.com/photon/components/message-bars.html */
 .message-box {
-  padding: 4px;
+  padding-right: 4px;
+  padding-left: 4px;
 
   border-radius: 4px;
 
@@ -120,9 +121,9 @@ a:active {
 }
 
 .icon-dismiss {
+  box-sizing: content-box;
   padding: 4px;
 
-  box-sizing: content-box;
   width: 16px;
   height: 16px;
 
--- a/src/common/common.css
+++ b/src/common/common.css
@@ -47,8 +47,8 @@ body {
   color: #ffffff;
   background-color: #d70022;
 }
-/* icon for error message */
-.message-box::before {
+/* icons for the message box */
+.message-box::before, .icon-dismiss {
   display: inline-block;
 
   /* fixed size */
@@ -70,6 +70,12 @@ body {
 }
 .success::before {
   background-image: url('/common/img/check.svg');
+}
+.icon-dismiss {
+  background-image: url('/common/img/close.svg');
+  position: absolute;
+  right: 10px;
+  cursor: pointer;
 }
 
 .info {
--- a/src/common/common.js
+++ b/src/common/common.js
@@ -423,7 +423,7 @@ const AddonSettings = (function () { // eslint-disable-line no-unused-vars
 const MessageHandler = (function () {// eslint-disable-line no-unused-vars
     const me = {};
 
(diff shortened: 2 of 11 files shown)


src/common/common.js
```diff
@@ -516,6 +516,17 @@ const MessageHandler = (function () {// eslint-disable-line no-unused-vars
...
         }
 
+        if (elMessage.classList.contains("dismissible")) {
+            // add an icon which dismisses the message if clicked
+            const iconDismiss = document.createElement("span");
+            iconDismiss.classList.add("icon-dismiss");
+            elMessage.appendChild(iconDismiss);
+
+            iconDismiss.addEventListener("click", () => {
+                elMessage.classList.add("invisible");
```
rugk: I think the event listener should send an event parameter, so better use that for hiding it. IMHO the scope is then much clearer. :smile:

src/common/common.css
```diff
@@ -71,6 +71,12 @@ body {
 .success::before {
   background-image: url('/common/img/check.svg');
 }
+.icon-dismiss {
+  background-image: url('/common/img/close.svg');
+  position: absolute;
+  right: 10px;
```
rugk: this "margin" should be 4px, see https://design.firefox.com/photon/components/message-bars.html#sizes-and-grid
but remember that the button/icon itself also already has a margin…

src/common/common.css
```diff
@@ -71,6 +71,12 @@ body {
 .success::before {
   background-image: url('/common/img/check.svg');
 }
+.icon-dismiss {
```
rugk: There are still some styles missing for hovering e.g.
AFAIK the correct styles are described as a [ghost](https://design.firefox.com/photon/components/buttons.html#ghost-3) button, but [this site](https://design.firefox.com/photon/components/message-bars.html#type-specific-style9) also shows some screenshots (when not hovered).

rugk: (I don't really know whether it should have a transparent background or really that grey one as specified)

src/options.html
```diff
@@ -4,7 +4,7 @@
 		<meta charset="utf-8">
 		<meta name="viewport" content="width=device-width, initial-scale=1.0">
 
-		<link rel="stylesheet" href="../common/common.css">
+		<link rel="stylesheet" href="common/common.css">
```
rugk: I doubt that works…

You can of course make the path an absolute one.

src/common/common.css
rugk: So [I've asked them about this](https://github.com/FirefoxUX/photon/issues/298).

src/options.html
ENT8R: But the path for the JS file does look like this `common/common.js`...

ENT8R:
> I think currently that is not really possible, is not it?

An hour ago it was not possible, but now it should work without any problems 😏

src/options.html
rugk: Well I thought just make it explicitly relative (like`/commons/…`) then, but well… yeah. Just strange why it worked then in the past.

Well… basically does not really matter.

src/common/common.js
```diff
@@ -516,14 +516,14 @@ const MessageHandler = (function () {// eslint-disable-line no-unused-vars
             elMessage.textContent = localizedString;
         }
 
-        if (elMessage.classList.contains("dismissible")) {
+        if (typeof args[1] === "boolean" && args[1]) {
```
rugk: This needs to be adjusted in the JSDoc, also in all functions, which call `showMessage`.

rugk: Also better use a const here for code style reasons (who knows what arg1 is, otherwise?). (or, better, [shift](https://github.com/rugk/offline-qr-code/pull/64/commits/5555d9ef24c2f50d68da0cf28bb9dc91973b654e#diff-ccdc4da75e7ac1061a8cc0851e3c7122R504) as I've done before…)

Also this codes likely needs to be moved some lines up, because [in line 515](https://github.com/rugk/offline-qr-code/pull/64/commits/5555d9ef24c2f50d68da0cf28bb9dc91973b654e#diff-ccdc4da75e7ac1061a8cc0851e3c7122R515) I already use _all_ other arguments for the translation of the message – so this would include the boolean…

src/common/common.js
```diff
@@ -516,14 +516,14 @@ const MessageHandler = (function () {// eslint-disable-line no-unused-vars
...
             // add an icon which dismisses the message if clicked
             const iconDismiss = document.createElement("span");
```
rugk: Okay, it creates this button – but it does not remove it (or at least make it invisible) when another message without "dismissable" set to true is shown.

As such maybe generally do it in another way: Add the HTML to the html file and only hide/unhide it when needed. This has two other advantages:
* if a message may never get a dismissable button, because of the way it is shown or so, it could just not be added to the HTML (such a case would be the Popup QR code error, IMHO). So in the JS you need to check for the existence of that element!
* That would also make customization easier.

rugk:
I tested it out and the right margin at border looks strange. I also made a thought mistake here reading the Photon guide. The 4px "margin" to the right, are actually the padding of the message box, which is already applied in the style – just because you use absolute positioning you are overwriting it there, as it seems.
The [margin around the icon](https://design.firefox.com/photon/components/message-bars.html#icon) is an _additional_ margin just around the icon, so in the end, we have 8px to the right. THis also looks okay, then:
grafik

However, it would be preferably if you could not use absolute positioning there anyway. Maybe [something like this](https://stackoverflow.com/questions/7693224/how-do-i-right-align-div-elements#12326152). Note the element is already in a flex, so you can happily use Flexbox there.

Generally, BTW, you can test out whether an auto-width (not 100%) looks better. I actually have no strong option on this (and Photon allows both styles based on vague "small/high width" definition), but currently I feel the "close" (dismiss) button is a bit too far away from the actual message.

src/common/common.css
rugk: Correction: See https://github.com/rugk/offline-qr-code/pull/64#issuecomment-385461872

rugk:
Also another problem: Based [on the screenshots](https://design.firefox.com/photon/components/message-bars.html#dismissible) I think the hover area needs to be actually include the 4px margin around the icon, i.e. be large enough, so that only the 4px padding of the message container is around the clickable area.

Currently the clickable area is just too small:
grafik

src/common/common.css
```diff
@@ -71,12 +71,22 @@ body {
...
+  right: 0px;
   cursor: pointer;
 }
+.icon-dismiss:hover {
+  background-color: rgba(12, 12, 13, 0.1);
+}
+.icon-dismiss:active {
+  background-color: rgba(12, 12, 13, 0.2);
+}
+.icon-dismiss:focus {
```
rugk: I just tested around a bit and the element is not even focusable by default. You need to add `tabindex="1"` to make it actually focusable. Then we can see the design.

Also, I think here yet again, a rounded border should be used (4px), although not explicitly specified in the Photon guide, but the screenshots look like this.

src/common/common.js
ENT8R: But all contents of the message box are replaced by setting `elMessage.textContent = localizedString;` [here](https://github.com/rugk/offline-qr-code/pull/64/files#diff-ccdc4da75e7ac1061a8cc0851e3c7122L516)...

ENT8R:
> Generally, BTW, you can test out whether an auto-width (not 100%) looks better

IMO this looks strange... I would also only search for a dismiss icon in the top right corner.

src/common/common.css
ENT8R: Isn't this documented [here](https://design.firefox.com/photon/components/buttons.html#default-3)? There they say "Corner Radius: `2px`"

rugk: Hmm, only for the default (big) button as far as I see, nothing is said for ghost buttons.
But yes, maybe use 2px, because if the other buttons have two pixels then this one may have, too.

ENT8R:
> I think the hover area needs to be actually include the 4px margin around the icon, i.e. be large enough, so that only the 4px padding of the message container is around the clickable area.

Do you have an idea how this can be achieved?

rugk:
I'll have a look, i.e. I take it over…

rugk:
TODO:
* maybe still move HTML code to all single items, so it does not need to be copied in JS? Would be useful if there are different messages with different (styles of) close buttons…
* maybe do not use border-box?
* maybe some fancy hide [animation](https://design.firefox.com/photon/motion/duration-and-easing.html), so it does not look so abrupt when hidden?

rugk:
Animation is done:
transitionhide

rugk:
So I moved the dismiss icon into the HTML source code, and basically that resulted in a larger refactoring than I thought. I also added some rules, which I did not thought of, to the contributing doc.

rugk:
~As for `box-sizing` I have no idea why it actually works…~

borderboxstrange

For some very strange reason [box-sizing](https://developer.mozilla.org/en-US/docs/Web/CSS/box-sizing) is actually set to `border-box` by default by some Firefox internal CSS:

That is contrary of what it is for websites, where 

So basically I think, @ENT8R, that was your issue when trying to use padding for the 4px difference there… Ugh…

As said [here](https://css-tricks.com/international-box-sizing-awareness-day/):
> You might think of it this way: with box-sizing: border-box the padding and border press their way inside the box rather than expand the box. 

Well… yeah, but unfortunately this is not what we want here, because padding with non-fixed height etc. is then totally useless as it is just calculated into the height and thus does not apply.

Now I understand why it is useful though. When I reset the `box-sizing` on the message-box, the 100% width is too large and it does not actually fit into the window, so it shows some scroll bars:
grafik

As I think the top and bottom padding is actually not intended, I changed the CSS a bit to match that.

…wtf…

I tried around it even more and the `box-sizing: border-box;` is maybe actually possible for two effects:
* that the QR code sometimes is cut off from the bottom.
* that my whole resizing to smaller sizes works, because you have still some pixel you can re-size (the padding more or less) without getting limited at some minimal size where it does not make the popup 5px smaller…

Sorry for the random thoughts, but this thing just totally caught me in surprise… 😵 It seems to cause both good and bad as you can see from the examples. Padding can go anywhere and be ignored, but you can have other good things…

rugk:
So I created a thread, [because it should not happen that I have to discover it this way](https://discourse.mozilla.org/t/add-ons-have-box-sizing-border-box-by-default/28359). That fundamental change is just not-documented anywhere.

ENT8R:
> @ENT8R BTW, just wanted to make you aware of one issue that slipped though…

Oh... Thank you for correcting this!

Issue #68, opened 2018-05-09
Use CSS variables


rugk:
We do have really too many hardcoded CSS values… (especially colors and stuff, which has assigned names in the Photon design guide)

So this needs to be changed.

See:
https://developer.mozilla.org/en-US/docs/Web/CSS/Using_CSS_variables

IMHO all variables can be created in the `common.css` to have them in one place even if we might not use them in all cases.

anirudh-modi:
Hey @rugk ,

I am looking for my first open source contribution, I would like to complete this task. 
Below is the list of css variable names which is assigned in Photon design guides.

https://firefoxux.github.io/design-tokens/photon-colors/photon-colors.css

Kindly let me know if these are the variables naming conventions which needs to be followed? 

And should I make the desired changes in `common.css` for declaration of variables and then go on to change color in files mentioned below :
* `common.css`
* `options.css`
*  `qrcode.css`

Thanks!
Anirudh

anirudh-modi:
Shall there be css variable for font size also??

https://design.firefox.com/photon/visuals/typography.html

There some of the naming defined here in the link above, however it seems as the styling depends on the os environment.

anirudh-modi:
Also looks like we need to create a copy of Mozilla Public License, v. 2.0. Correct me if I am wrong about it.

rugk:
Hi @anirudh-modi,
thanks for your interest in contributing to this project! :tada: :+1: 
I hope you'll like this project and can enjoy hacking on it… :smiley:

> Kindly let me know if these are the variables naming conventions which needs to be followed?

Just use the style, the Firefox photon guide uses. So we don't need any conversion of the variable names here, that makes copy and past easier. :wink: 

> And should I make the desired changes in common.css for declaration of variables and then go on to change color in files mentioned below

Indeed, that's what I imagine. :+1: 

> Shall there be css variable for font size also??

Good question, but I don't think so. Firstly currently I use some custom sizes, which just match the text size the Firefox add-on settings page uses, so it may not actually be covered in that guide.
Secondly, the Photon guide says it is different from OS and then you have all these variants (weight and stuff…), so i think it is not worth it currently.

anirudh-modi:
Hi @rugk,

Thanks alot for the information, I shall assign this task to myself and do the needful.

rugk:
FYI you can link to lines on GitHub, but I guess you copied them from a text editor, so…

> ` background-color: #d70022ee;`

[...]

PR #80, opened 2018-05-14
Css variable

anirudh-modi:
Hi @rugk 

I have made the changes as discussed for the issue #68 . Kindly let me know if anything is incorrect from my side. Thanks, for giving me a chance to contribute!!

### DIFF
--- a/src/common/common.css
+++ b/src/common/common.css
@@ -1,7 +1,3 @@
-/* This Source Code Form is subject to the terms of the Mozilla Public
- * License, v. 2.0. If a copy of the MPL was not distributed with this
- * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
-
 /* Photon Colors CSS Variables v3.2.0 */
 
 :root {
--- a/src/common/common.css
+++ b/src/common/common.css
@@ -1,7 +1,94 @@
 /* Photon Colors CSS Variables v3.2.0 */
+/* https://firefoxux.github.io/design-tokens/photon-colors/photon-colors.css */
 
 :root {
   --magenta-50: #ff1ad9;
+  --magenta-60: #ed00b5;
+  --magenta-70: #b5007f;
+  --magenta-80: #7d004f;
+  --magenta-90: #440027;
+
+  --purple-30: #c069ff;
+  --purple-40: #ad3bff;
+  --purple-50: #9400ff;
+  --purple-60: #8000d7;
+  --purple-70: #6200a4;
+  --purple-80: #440071;
+  --purple-90: #25003e;
+
+  --blue-40: #45a1ff;
+  --blue-50: #0a84ff;
+  --blue-50-a30: rgba(10, 132, 255, 0.3);
+  --blue-60: #0060df;
+  --blue-70: #003eaa;
+  --blue-80: #002275;
+  --blue-90: #000f40;
+
+  --teal-50: #00feff;
+  --teal-60: #00c8d7;
+  --teal-70: #008ea4;
+  --teal-80: #005a71;
+  --teal-90: #002d3e;
+
+  --green-50: #30e60b;
+  --green-60: #12bc00;
+  --green-70: #058b00;
+  --green-80: #006504;
+  --green-90: #003706;
+
+  --yellow-50: #ffe900;
+  --yellow-60: #d7b600;
+  --yellow-70: #a47f00;
+  --yellow-80: #715100;
+  --yellow-90: #3e2800;
+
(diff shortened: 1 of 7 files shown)


src/common/common.css
```diff
@@ -1,24 +1,49 @@
+/* Photon Colors CSS Variables v3.2.0
+   https://firefoxux.github.io/design-tokens/photon-colors/photon-colors.css */
+:root {
+  --blue-50: #0a84ff;
+  --blue-50-a30: rgba(10, 132, 255, 0.3);
```
rugk: > Since the base color was defined in hex I could not use rgba to generate the error background witha transparency, 

So as you modify the copy here anyway, why not convert the orginal colors into rgb format? (But would it actually work anyway?) As you modified the code after coping from the source anyway, one cannot easily copy & paste it anyway?

anirudh-modi: Yes If I go ahead and change the base color to rgb format it would work.  

It is just to maintain consistency within code, we have to convert the base hex code into rgb and then use it. I will go ahead and convert the base color into rgb.

anirudh-modi:
> Generally maybe do consider moving the stuff you add as an addition to the original source into a new :root block, so one can copy & paste the original source when updating is needed. (and have the own modifications/additions separate)

My apology but I didnt quite get what you were trying say. 

> If you want, you may also add it as a new style sheet and just include it in the HTML files. (Considering the licensing stuff mentioned below, I would maybe also prefer this.)

I will create a new css file `default.css` where all the future declaration variable can go, and include it the respective `html` files.

anirudh-modi:
> And you also need to adjust the license file in this repo to indicate that this code . Also better include the "his Source Code Form is subject to the terms of the Mozilla Public[…]" stuff as a comment here, but of course it should be clear this only applies to the first block/only this code part.

For the above I will include the comment in the newly created `default.css` specifed the below : 
```
<!-- This Source Code Form is subject to the terms of the Mozilla Public
   - License, v. 2.0. If a copy of the MPL was not distributed with this
   - file, You can obtain one at http://mozilla.org/MPL/2.0/. -->
```

anirudh-modi:
My mistake I accidently closed it.

anirudh-modi:
@rugk I have made the changes in the latest commit, as suggested by you..

However I have still not added anything in `License.md` file. So, about that, I can see that there is already a section for [Mozilla Photo Icons](https://github.com/rugk/offline-qr-code/blob/master/LICENSE.md#mozilla-photon-icons) and then the license for the same, Should I change the header title from `Mozilla Photon icons` to `Mozilla Photon icons and colors` and add the link to the `default.css` within the same section?

rugk:
Looks good, I may change some minor stuff, but basically good to merge, IMHO. :smiley:

anirudh-modi:
Thanks, for being patient!! And approving the request!!

anirudh-modi:
So, I had forked your repo and I was making changes in my own branch, I have merged my branch to the master branch of my forked repo.

What to do next? I mean Do i have to create a new PR to merge my master to yours?

rugk:
So as you may see I have added a new commit to your branch (https://github.com/rugk/offline-qr-code/pull/80/commits/8dc7262bd8e39204cf07f85e659c2b75d4a7870d). While testing out your changes, I've discovered two things you should know (and I already fixed them while I was at it, BTW):
* first, indentation for CSS files is 2 spaces in this project; we also support a handy tool to help you here; see https://github.com/rugk/offline-qr-code/blob/master/CONTRIBUTING.md#coding-guidelines
* Secondly, I saw that the way of adding transparency via `rgba(--var, 0.x`) doe snot actually work (at least not in Firefox 60). See also https://stackoverflow.com/questions/29591465/use-css-variables-with-rgba-for-gradient-transparency#29591536, so I think there is no nice way for that. This is how the success message (when you reset the options) in the options e.g. should look like when the dismiss button is hovered (grey transparency):
    var
    It actually did not look like that in my tests.

    As such, I have also converted the colors back to hex, because having them in rgb format now does not help anymore.

anirudh-modi:
Oh my sincere apologies, from next time onwards, I will make sure that I test things out a little more properly. 

Thank you for being so helpful, patient and allowing me to make first PR to your repo! 

I will see if there is any other issue which can be completed.

anirudh-modi:
I hope you didn't have much trouble in collaborating with me.

anirudh-modi:
Ok. My mistake, I actually skipped testing var in rgba, and sort of assumed that it might work. Would keep it in mind next time.

Issue #20, opened 2018-04-16
Warning for inverted QR codes (when background color is dark and foreground light)


rugk:
Similar to https://github.com/rugk/offline-qr-code/issues/19 there should be warning. But this time, it should just say that you can only scan such inverted Qr codes with some QR code readers (e.g. Android "Barcode reader" with a special option).

Maybe also create a wiki page (or FAQ?), for such a thing and explain how to scan it.

Depends on https://github.com/rugk/offline-qr-code/issues/19

LMLB:
I have separated the luminance calculation into it's own function in [the gist I posted in that issue](https://gist.github.com/LMLB/47cb7997f9e7e4cf1f1052f80f93b195).

rugk:
Great, but did you see that @ENT8R already created an own PR, which includes the contrast (and thus, as it seems luminance) calculation: https://github.com/rugk/offline-qr-code/pull/86/files#diff-91ba5f4df12e1b6b0629662164b3f14eR245?
(Although I'd think for code style reasons, your version really looks better, so @ENT8R you use that. Of course, you can also give @LMLB push access to that branch.)

But as for this issue (i.e. "finding out which color is dark/light") I don't think we need a new function for that luminance thing. Could not the contrast ratio calculation just return `-3` or `+3` (or speaking in general terms `-x`/`+x`) depending on in which order the colors are compared?
But if it is cleaner or easier with that additional functions, so be it. :smile:

rugk:
Sounds good…

rugk:
BTW, @LMLB, – I know, I am a bit late –, but if you wanna do a PR, feel free to do so.

But I've got some other color issues in this repo, too. :smile:

rugk:
I am assuming you will do no PR, @LMLB.

PR #86, opened 2018-05-19
Add warning for low contrast and equal colors

### LINKED ISSUE #19: Warning when settings are low-contrast for QR codes
There should be a warning setting when the contrast of the background/foreground colors is too low. It should say that scanning QR codes with such a low contrast may not be possible.

I guess as one needs to use some color comparison, this issue could be a bit complicated and maybe a new library needs to be introduced. So sharing your ideas before making a PR would be appreciated.

Requires:
* the warning error message is not completely, so this would need. As with the others, use FF Photon icons.
* warning message should be shown immediately, but should not block saving the information; i.e. just show it, but
* warning message may be dismissable or not (I'd rather prefer not, as the warning is obviously always applicable.) If not, it must be shown again when the settings are opened another time, of course.
   Or you could save the dismissable status, but that would be some more work.

### LINKED ISSUE #83: Implement (styling of) warning message
Currently I had no use for showing a warning message, but that may change.

In any case, I think, it is useful for the future.

The JS stuff is already implemented as far as I see (but test it, nevertheless!). It uses the element with id `messageWarning` for warning messages.

What needs to be done is:
* look at how other messages are implemented/added and do it the same for the warning message
* adhere to the design guide for message bars: https://design.firefox.com/photon/components/message-bars.html
* test that thing with some warning message – maybe you even find a place, where it could be useful to add one

ENT8R:
This PR fixes #19 
**Edit:** it also fixes #83 
I'm not sure what should be the best contrast ratio (currently it is 4 because the tests I've done with https://dequeuniversity.com/rules/axe/2.2/color-contrast returned that 4 is OK if the text is not too small...)
The amount can be changed here: 
https://github.com/ENT8R/offline-qr-code/blob/819d68d403137df93eda018b10553e35e1e010fd/src/options.js#L217

screenshot 4
screenshot 3


rugk:
Great PR! From a first look at the code it looks good.

And it seems you also fixed https://github.com/rugk/offline-qr-code/issues/83 (you may add it to the PR description here, so it is automatically closed).

As for the error message: good point to show an error if the color is exactly the same. However, you may show it even earlier when it is not at all readable by a qr code scanner at all even if some text would still be visible to the human eye with that contrast.

So could you maybe do some tests with some qr code scanners to see at which contrast ratio a qr code can still be scanned. (Or maybe you find an official document/spec, which suggests a minimum contrast)

So the levels should maybe be like:
* warning: may not be scannable by, say, >50% of most qr code scanners
* errors: certainly not scannable even with most good mobile qr code scanners

BTW, if you want, you can make use of the new "action buttons" I've implemented for messages. Maybe reset the last changed color? Or automatically calculate a good color, like the exact contrast color? That would make the error messages even Moor helpful.

Remember there is also https://github.com/rugk/offline-qr-code/issues/20, which is very similar and may be accomplished with the same code or just a minor change, or so.

src/common/common.js
```diff
@@ -732,7 +732,7 @@ const MessageHandler = (function () {// eslint-disable-line no-unused-vars
         if (typeof args[0] === "boolean") {
             isDismissable = args.shift();
         }
-        if (typeof args[0] !== undefined && args[0].text !== undefined && args[0].action !== undefined) {
+        if (typeof args[0] !== "undefined" && args[0].text !== "undefined" && args[0].action !== "undefined") {
```
rugk: In ES6 it is no longer needed to compare the typeof value to the string of `undefined` explicitly. See https://stackoverflow.com/questions/34596489/es2015-2016-way-of-typeof-varname-undefined#34596516

src/common/common.js
```diff
@@ -770,7 +770,7 @@ const MessageHandler = (function () {// eslint-disable-line no-unused-vars
 
             elActionButton.textContent = browser.i18n.getMessage(actionButton.text) || actionButton.text;
             elActionButton.classList.remove("invisible");
-        } else {
+        } else if (elActionButton) {
```
rugk: Oh, good point! 

Dismiss icon (https://github.com/rugk/offline-qr-code/pull/86/files#diff-ccdc4da75e7ac1061a8cc0851e3c7122R751) needs the same adjustments. (but I can also do them by myself)

src/common/common.js
ENT8R: Mmh... But it gave me an error and when I printed `typeof args[0] !== undefined` it returned `false` even if the array was empty. And `typeof args[0] !== "undefined"` returned `true`...

ENT8R:
> So could you maybe do some tests with some qr code scanners to see at which contrast ratio a qr code can still be scanned.

I just tested it with Zxing and it scanned even codes with a contrast ratio of below 2 (1 is the lowest possible value). The only problem were inverted codes (which could not be scanned at all below a contrast ratio of 5) but that should be covered with #20... I will do now some more tests with other applications.

src/common/common.js
rugk: Firefox 60:

grafik

rugk: And and yeah… it's wrong. As https://stackoverflow.com/questions/34596489/es2015-2016-way-of-typeof-varname-undefined#34596516 says, we may better not use that "typeof" for it anyway. In ES6 one can just check whether it is undefined directly.
So better not use `typeof`.

rugk:
> The only problem were inverted codes (which could not be scanned at all below a contrast ratio of 5) but that should be covered with #20...

Indeed, also Zxing ("Barcode Reader") is one of the few apps, that actually support scanning inverted QR codes. You have to enable it in the options.

src/common/common.js
ENT8R: Alright. I will change that.

src/options.js
```diff
@@ -200,11 +202,81 @@ const OptionHandler = (function () {
...
+    }
+
+    /**
+     * Converts a hex color string to RGB.
+     *
+     * @name   OptionHandler.hexToRgb
+     * @function
+     * @private
+     * @param  {string} hex
+     * @returns {Array}
```
rugk: this function can also return `null` as it seems…

src/options.js
```diff
@@ -200,11 +202,81 @@ const OptionHandler = (function () {
...
+     */
+    function hexToRgb(hex) {
+        const result = /^#?([a-f\d]{2})([a-f\d]{2})([a-f\d]{2})$/i.exec(hex);
+        return result ? [
```
rugk: Uh, splitting the conditional operator on multiple lines? Better use a traditional `if` here.

ENT8R:
> Or maybe you find an official document/spec, which suggests a minimum contrast

I found no official document. All I found just said that the contrast should not be too low 😄 
But based on my tests with 3 other apps besides Zxing I will show the warning when the contrast ratio is below 6 and the error if the ratio is below 3. But if somebody has other experiences, those values can be changed easily...

> Or automatically calculate a good color, like the exact contrast color?

Do you mean the [opposite color](https://stackoverflow.com/a/35970186/8575129) of the QR code (or the background)?

rugk:
> I will show the warning when the contrast ratio is below 6 and the error if the ratio is below 3.

Okay, so I need some practical examples for that…

This is 6:
grafik

grafik

grafik

So for the same color examples it may be okay for the user, but the QR code scanner is what matters, so I also tested some colors…

Below 2 it stopped for most of my QR code scanners:

grafik
grafik

Around 2 it differed… Funnily particularly ZXing was bad at it. Another scanner took some seconds…


grafik

Around 3 (image above) even ZXing got it, but took some seconds…

Around 4 all got it, but actually the perceieved contrast one would imagine is quite good– and it looks great from a design point of view:
grafik


However, I had a quite good light conditions (no darker screen) here, this may also come into effect.

----

So I think 6 is way too much to require. However, it also depends on the light source…
In general we might not want to limit that contrast ratio to integers anyway.

So maybe we need a less scary message. So can you:
* make `contrast` return a float
* introduce a simple "info message" (no warning, just a small note), for say – lower than 4.5 (that's WCAG 2.0 level AA for small text, BTW)
* a warning for smaller than 3:1 (that's WCAG 2.0 level AA for large text, BTW)
* and an error for <2

Otherwise please explain the tests you did again, I just did some few I just listed. Of course, if we want a realistic result we need more tests. Please explain the results you had in more detail, maybe we can adjust the number again or my tests were flawed (did not test low screen brightness e.g.).
I want to avoid showing too much errors and prevent some nice-looking QR codes, but of course they should stil be scannable by the majority of QR code scanners in most of the situations.

-----

> Do you mean the opposite color of the QR code (or the background)?

Yeah.

rugk:
BTW this color stuff get's much if you want to also implement support for opposite color there. So maybe you could split that color stuff (only that color caluclation/etc.) into it's own module. (Could be in commons.js or in options.js – does not really matter, as we currently won't use it anywhere else than in the options, but it will be restructured into `import` modules of Firefox 60 anyway, soon.)

ENT8R:
It was a little bit more work than I thought but now there is a small button next to each message which (when clicked) calculates the complementary color of the QR code and sets the calculated color as the new background of the code automatically.

> Otherwise please explain the tests you did again, I just did some few I just listed.

I just donwloaded 3 different QR code scanner apps and tried some random settings while seeing the contrast ratio value in the console. I think your tests were more extensive...

src/_locales/en/messages.json
```diff
@@ -62,14 +62,22 @@
...
+  "lowContrastRatioWarning": {
+    "message": "The contrast ratio between the QR code color and the background might be too low to be recognized by the most QR code readers",
+    "desription": "The warning shown when the contrast ratio is too low"
+  },
+  "lowContrastRatioError": {
+    "message": "The background and the foreground color are too similar.",
     "description": "The message shown when the color for the QR code and the background are equal"
   },
+  "messageAutoSelectColorButton": {
+    "message": "Calculate best contrast",
```
rugk: MAybe you can try to shorten this a bit?

Instead of calculate maybe just "Use". (That the JS has to calculate this is a technical detail the user does not even want to know. :smile:)
And instead of contrast maybe contrasting color? Or something else like "best color"? Remember it does not matter to be technically 100% accurate here, the user should just know what happens when they click the button.

Theoretically it could e.g. also be okay for the user to get some color variation (i.e. not 100%ly the contrast color, but a color, which has enough contrast, so the message goes away.)

src/options.js
```diff
@@ -200,6 +203,47 @@ const OptionHandler = (function () {
...
+                        Logger.logError("could not save option", option, ": ", error);
+                        MessageHandler.showError("couldNotSaveOption", true);
+                    }).finally(() => {
+                        applyOptionLive();
+                        elQrBackgroundColor.value = invertedColor;
+                    });
+                }
+            };
+
+            // breakpoints: https://github.com/rugk/offline-qr-code/pull/86#issuecomment-390426286
```
rugk: Oh and once we are at it. You can link to my comment, yes, but please still add comments for the colors, i.e. the accessibility guidelines I mentioned.

Or even better remember the "no magic numbers" rules. So put these numbers into constants. As they are color stuff not (only) related to the options handler, the best would be to use these as constants in the `Colors` module, i.e. make them "public constants". (add to `me` in that module)

ENT8R:
> Apart from that, there is one bigger: If you do it in the wrong way, you can display multiple messages:

Really? How does that work? I can't find the issue in the code currently 😄...

src/common/common.js
```diff
@@ -1372,6 +1372,110 @@ const RandomTips = (function () {// eslint-disable-line no-unused-vars
     return me;
 })();
 
+const Colors = (() => { // eslint-disable-line no-unused-vars
+    const me = {};
+
+    me.ContrastBreakpoints = {
+        A: 2,
```
rugk: No my numbers were not all the A/AA/AAA things…

But I'll change that by myself.

rugk:
You just change the colors again and again and thus "fix" them. When one message is shown, it does not go away unless you got no message at all… :smile: 

BTW I changed the constant there. And I could not find any contrast ratio definition for WCAG level A, I guess they start at level AA for contrasts.

rugk:
Any ETA on this?

ENT8R:
> Any ETA on this?

I think I fixed the issue where multiple messages where shown. Is there still something to do? I could not find anything...

rugk:
Merged in https://github.com/rugk/offline-qr-code/commit/313a5f6c61b7363ffc81929bf33aa4651a050666

PR #106, opened 2018-06-23
Fix remember last size

### LINKED ISSUE #102: Remember last size does work as expected
I changed the QR code size option to 'remember the last size' but it could not restore to the previous size on next launch.
Windows 10
Firefox 60.0.2

AmauriAires:
- Fixed #102 by adding variable initialClick.

### DIFF
--- a/src/popup/qrcode.js
+++ b/src/popup/qrcode.js
@@ -450,7 +450,6 @@ const UserInterface = (function () {
     // default/last size
     let qrLastSize = 200;
     let qrCodeSizeOption = {};
-    let initialClick = false;
 
     /**
      * Hide QR code and show placeholder instead.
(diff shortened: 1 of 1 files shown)


AmauriAires:
I have tried many things with the MutationObserver without success. I will leave this solution (or workaround) here for now. : )

rugk:
Hmm okay, did you try the workaround with `lateInit` I mentioned earlier? `lateInit` runs later than `init`, but not as late as the loading thing, which did not work. It may be that – to make the observer work – you should still create it in init, but only start in lateInit. (or something similar?)

rugk:
So I've looked into it and found a solution, which is a bit cleaner, but works more or less in the same way: https://github.com/rugk/offline-qr-code/commit/353192e67bac5872f1b3a9e016e1b6c679672723

Thanks for finding the actual problem/cause though.

Issue #89, opened 2018-05-19
Should show error if tabs cannot be retrieved


rugk:
This e.g. happens like this (STR):
1. Open a new tab and load some website.
2. While it loads quickly open/click on the QR code popup.

What happens:
It seems it cannot get the tabs and just shows a loading message.

What should happen:
It should show an error instead.

I guess the fix should be quite easy, but involves some stuff with Promises, so you need knowledge in that. Reproducing may be the hardest of this issue, as it requires some good timing… :wink: 

(also only reproduced with an old version; some stuff has changed since that, maybe it is actually already fixed; in this case, it just requires a conformation that this is fixed.)

krishnan2098:
I would love to help but i dont know how to start because i am a newbie

rugk:
Well… first of all you need to reproduce the issue. Can you do so with the current version?

If so follow [this guide](https://github.com/rugk/offline-qr-code/blob/master/CONTRIBUTING.md#getting-started) to test it with the latest version. Still reproducible? Then you can try to find the root cause. Cannot reproduce? Then it has been fixed in the mean time and you can jsut comment here and say it seems it has been fixed and this can be closed.

If you need to debug it, look somewhere around here:
https://github.com/rugk/offline-qr-code/blob/master/src/popup/qrcode.js#L1092

This is the line that should actually display the error. Maybe the code does not reach that part? Best thing would be to try using [the debugger](https://mozilladevelopers.github.io/playground/debugger/) to find out what happens in this case and why it does not show this error.

krishnan2098:
It seems it has been fixed as I have cloned the latest version in my local machine and tested it a few times and the wasn't any error, is there anything else where I can help you?

rugk:
> and the wasn't any error,

Ehm… well, this issue would be fixed if _there is_ an error. Maybe read the STR again (https://github.com/rugk/offline-qr-code/issues/89#issue-324654755), this issue here is about showing an error, when it

krishnan2098:
### This is what it shows. There isn't any error message or warning about it.

![screenshot from 2018-05-26 18-55-43](https://user-images.githubusercontent.com/26490172/40576586-9b51ac78-6116-11e8-812d-7a899cb64139.png)

rugk:
And the QR code reads `undefined`. Yes, that is definitely a bug… it should show an error.

However, how did you trigger it? You do not seem to have opened it while the Google search page was shown.

krishnan2098:
[...]

PR #108, opened 2018-07-10
Added retry mechanism for tabs

### LINKED ISSUE #98: Recover from still loading tab
After the fix of https://github.com/rugk/offline-qr-code/issues/89 it now just shows an error if the tab could not be got.

It would be better though, if it could at least retry (with a timeout!) getting the tab URL when the tab is still loading. This is likely nicer than just showing an error, just because of some internal errors…

How-to: https://github.com/mdn/webextensions-examples/blob/8694314d5a328616f028bb71e4d6bbd352375440/session-state/background.js#L37-L51

OogieBoogieInJSON:
Fixes #98 

I've tried with the onUpdate listener but aparently the behavior is unpredictable. When the event is triggered with status being complete all urls are missing but the API docs say they should be there.
I ended up with this implementation but still if you are super quick after pressing enter to navigate to that url and then open the extension, the tabs query will not have the urls in there. By super quick, I mean under a second.


src/popup/modules/QrLib/retryPromise.js
```diff
@@ -0,0 +1,15 @@
+/**
+ * Try fn when the return promise is rejected
+ * @param {Function} fn function to retry
+ * @param {number} ms retry delay in ms
+ * @returns {Promise}
+ */
+const retryPromise = (fn, ms) => new Promise(resolve => {
```
rugk: Wow, pretty clever stuff here and functional programming as it should be. ☺️

Just some minor notes:
* do not abbreviate the var names so much, rather call it "delay"/"time" or so, same with "function". We have enough space here. :wink: 
* you can (and may :wink:) use the usual `function(` syntax here instead of arrow functions and a variable assignment. Just so it looks less creepy to ones knew to these stuff… :wink: 
* everything else, great! Maybe move it to the `commons` dir though, as it is not limited to the popup functionality…
   and, yeash, until now I did not use default exports. Maybe just put it into the "HelperFunctions.js" file there – or how it is called – …

src/popup/modules/InitQrCode.js
```diff
@@ -85,3 +85,38 @@ export const initiationProcess = Promise.all([qrCreatorInit, userInterfaceInit])
...
+ * Will reject if url is not defined.
+ * @returns {Promise}
+ */
+async function getsTabWithValidUrl() {
+    let retryCount = 0;
+    const maxRetries = 20;
+    const delay = 300;
+
+    try {
+        return await retryPromise(async () => {
```
rugk: Hmm, is the `await` here really needed?

src/popup/modules/InitQrCode.js
```diff
@@ -85,3 +85,38 @@ export const initiationProcess = Promise.all([qrCreatorInit, userInterfaceInit])
...
+
+/**
+ * Queries Tabs API for 20 times with a delay of 300ms until the tabs have a defined url.
+ * Will reject if url is not defined.
+ * @returns {Promise}
+ */
+async function getsTabWithValidUrl() {
+    let retryCount = 0;
+    const maxRetries = 20;
+    const delay = 300;
```
rugk: As these are constants and literals you may write them `CONSTANT_STYLE`. But here they are not in a module scope, but in a function (as it is explained in the contributing doc), so I am not sure here… May think about this later… :thinking:

rugk:
Hi @OogieBoogieInJSON,

Actually that PR is quite impressing from your functional code style. :smile: 
As far as I see, it just retries getting the tab though…

So:
> When the event is triggered with status being complete all urls are missing but the API docs say they should be there.

So this definitively rather sounds like a browser bug to me. Are you sure it is happening like this?
Not only the docs say so, but also this [session-state example I've linked previously](https://github.com/mdn/webextensions-examples/blob/8694314d5a328616f028bb71e4d6bbd352375440/session-state/background.js#L37-L51) seems to use that behaviour. Okay, it does not access these properties… 😐

---

Okay, I guess what the issue here is: 
Do have a look [at how I saved](https://github.com/rugk/offline-qr-code/commit/ff666cd90fd51951a232971d0eb1b80fbf9f229f) the previous issue (https://github.com/rugk/offline-qr-code/issues/89). So the promise may resolve, but have an undefined URL.

Still wonder though, whether this is expected (i.e. it maybe returns the URL of "empty tab"??) or a browser bug.

OogieBoogieInJSON:
What do you mean with with #89 ? That I should call MessageHandle.hideError() ?

src/popup/modules/InitQrCode.js
OogieBoogieInJSON: I won't consider these application level constants, they are just used inside this function. I could have just pass the numbers to retryPromise without assining them first to a variable.

src/popup/modules/InitQrCode.js
OogieBoogieInJSON: yes of course, retryPromise returns a promise, it has to wait for it until it's resolved/rejected

src/popup/modules/QrLib/retryPromise.js
OogieBoogieInJSON: Thank you, i'm a big fan of simplicity and composability.
I'll address these points and make declarations more descriptive.

src/popup/modules/InitQrCode.js
rugk: Yeah, I had the same thought. So I guess it is okay.

src/popup/modules/InitQrCode.js
rugk: Promises are there for, so they can run asynchronously and stuff does _not_ have to wait though.
So I'd bet that you do not need the await here. 

All the waiting it only needs should be done by setTimeout and this here can just be returned. Promises are asynchronous – taht's the point, so I guess it should work.

The code does not have to"wait" explicitly to evaluate the Promise result, it "jumps back" to the last execution point, if the result is there. Or am I missing something here?

src/popup/modules/InitQrCode.js
```diff
@@ -85,3 +85,38 @@ export const initiationProcess = Promise.all([qrCreatorInit, userInterfaceInit])
...
+
+                const tabs = await queryBrowserTabs();
+
+                if (tabs && tabs[0].url) {
+                    return tabs;
+                } else {
+                    throw new Error("Url not found.");
+                }
+            } catch (error) {
+                throw error;
```
rugk: This catches and directly throws the error again?? I guess this is not needed. So we can have one less try block here… Same below. Thrown errors bubble "through" until they are catched. So there is no need to catch them, just to throw them directly thereafter. (They work like exceptions.)

src/popup/modules/InitQrCode.js
```diff
@@ -85,3 +85,38 @@ export const initiationProcess = Promise.all([qrCreatorInit, userInterfaceInit])
...
+                throw new Error("No tabs with url.");
+            }
+
+            try {
+                retryCount++;
+
+                const tabs = await queryBrowserTabs();
+
+                if (tabs && tabs[0].url) {
+                    return tabs;
```
rugk: Hmm, the function says `getsTabWithValidUrl`, so it says it returns one tab, but here it returns all of them.

So I guess you need some small re-factoring to adjust that behaviour. I'd prefer if it really just returns one tab. Of course, you may adjust/delete some functions of the `QrCreator` then.

rugk:
> What do you mean with with #89 ? That I should call MessageHandle.hideError() ?

No… wait… In https://github.com/rugk/offline-qr-code/pull/108/files#diff-9fb6283c7183ac6eeb9e365bb2038026R110 you already check for `tab.url == undefined` basically, so what you say is already done, mhh… :thinking: 

---

Anyway, let's re-consider the `onUpdate` stuff. It must work – somehow. We cannot just say "does not work" and ignore it. In any case, we could first wait for `onUpdate` and then trigger all these custom re-trying you provide in this PR.

src/popup/modules/InitQrCode.js
```diff
@@ -5,12 +5,13 @@ import * as MessageHandler from "/common/modules/MessageHandler.js";
...
 import * as BrowserCommunication from "./BrowserCommunication.js";
 import * as UserInterface from "./UserInterface.js";
+import { retryPromise } from "../../common/modules/HelperFunctions.js";
 
 /* globals */
 export let initCompleted = false;
 
 // init modules
-const queryBrowserTabs = browser.tabs.query({active: true, currentWindow: true});
+const queryBrowserTabs = () => browser.tabs.query({active: true, currentWindow: true});
```
rugk: Actually, with your change, this now does not need to be a the global scope anymore. This variable is only used once, so you do not need to save it in a global var. :smile:

src/common/modules/HelperFunctions.js
```diff
@@ -10,3 +10,19 @@
 export function objectIsEmpty(obj) { // eslint-disable-line no-unused-vars
     return Object.keys(obj).length === 0 && obj.constructor === Object;
 }
+
+/**
+ * Try fn when the return promise is rejected
```
rugk: `fn` abbreviations here. :smile: 
Also, if you are at it, also always add an empty line before the `@` params. This is how it's done in other JSDOC in this project, so just stay consistent here.

rugk:
**Edit:** See edit below, I totally misunderstood `onUpdate`…

----

> IMO doing both becomes superfluous because then I have to sync one event listener with two promises which is going to end either in some nice spagetti or more boilerplate.

~~First I don't think it will end so bad, because your code was quite elegant, so I am sure you'll find a nice way to do so.~~ :wink:
~~My approach would be to first wait for the onUpdate being "completed" and then start your stuff with potentially re-trying all stuff. So not really sync anything, just do it synchronously. AFAIK the URL is never available before the "onUpdate completed" event is sent, so we do not need to try getting it before, is not it?~~

~~But about the "superfluous"… You've said previously about your implementation: ~~

> but still if you are super quick after pressing enter to navigate to that url and then open the extension

~~So I thought using both these things would fix it:~~
* ~~`onUpdate` -> completed makes sure a minimal amount of time has elapsed and the stuff should actually be ready~~
* ~~your re-trying then may even be a shorter time period and ~~

~~I guess this behaviour is just more logical. Just re-trying stuff is a kind of "brute-force way" and still kinda like a workaround.~~

Also I'd still definitively ask ([here e.g.](https://discourse.mozilla.org/c/add-ons/development)) whether/how it is expected that sometimes the URl is not there even after it is completed. Naturally I'll still try to reproduce it before. Of course, if you want, you can start asking it there.

But maybe better explain the issue of this "super quick" anyway a bit more in detail, because actually, the aim of this PR is to actually fix that. If you are too fast, the add-on should just re-try some time later whether the URl is now available. And as far as I see, this should work.

**Edit:** Oh, well… no. Now I get it. `onUpdate` is of course only triggered if a new tab is created/status changes or so. So this could well be more complicated.
So forget what I've said before. Maybe it is really the only way by just re-trying.

rugk:
What you may still do:
* `onUpdate` is still supposed to return an URL. This does not happen? I guess it's a bug -> report or ask somewhere, as I've described

OogieBoogieInJSON:
> Because, as said, I don't understand it. As that's the issue this PR is supposed to fix.

Well, IMO it's unfixable at the moment, we can close the pull request without merging. But it's still better than before when until the page did more seconds to load and after that opening the extension the URL wouldn't be there.

src/popup/modules/InitQrCode.js
OogieBoogieInJSON: True that, I moved it inside the function

src/common/modules/HelperFunctions.js
OogieBoogieInJSON: i've change the description, sometimes im so bad at jsdocs.

src/popup/modules/InitQrCode.js
```diff
@@ -92,6 +91,7 @@ export const initiationProcess = Promise.all([qrCreatorInit, userInterfaceInit])
  * @returns {Promise}
  */
 function getTabWithValidUrl() {
+    const queryBrowserTabs = () => browser.tabs.query({active: true, currentWindow: true});
```
rugk: Okay, but actually you do not need to save it in any variable...
You only call it once in line 106.

src/popup/modules/InitQrCode.js
```diff
@@ -92,6 +91,7 @@ export const initiationProcess = Promise.all([qrCreatorInit, userInterfaceInit])
  * @returns {Promise}
```
rugk: here is also an empty line missing, ... also jsdoc

rugk:
>Well, IMO it's unfixable at the moment

But why? When you can't get it a 001 ms after loading the site, it should _always_ be able to get the URL -  say - after 1000ms after loading the tab. At some point in time, it must have loaded the tab (and it's URL).

src/common/modules/HelperFunctions.js
```diff
@@ -10,3 +10,20 @@
 export function objectIsEmpty(obj) { // eslint-disable-line no-unused-vars
     return Object.keys(obj).length === 0 && obj.constructor === Object;
 }
+
+/**
+ * Recall the function with a delay when the returned promise is rejected
+ *
+ * @param {Function} fn function to retry
```
rugk: `fn`...

src/popup/modules/InitQrCode.js
```diff
@@ -5,12 +5,12 @@ import * as MessageHandler from "/common/modules/MessageHandler.js";
 import * as QrCreator from "./QrCreator.js";
 import * as BrowserCommunication from "./BrowserCommunication.js";
 import * as UserInterface from "./UserInterface.js";
+import { retryPromise } from "../../common/modules/HelperFunctions.js";
```
rugk: for common modules, you can use an absolute path (because that `..` does not bring any information value), just have a look how I did it in other modules.

OogieBoogieInJSON: Hmm, is this part of the modules spec?

rugk: Well... actually you can use path (without variables or so) in ES6 modules. That `/` is the root dir in WebExtensions, is actually a feature of WebExtensions (because you obviously have no website root dir).

OogieBoogieInJSON:
@rugk you are right. Sadly it doesn't. The onUpdate seems to be broken. So, how would you like to proceed? Merge this one to fix things temporarily and I'll open a ticket for Mozilla and ask there?

rugk:
> @rugk you are right.

With what? I actually asked this as question to you.
Let me reiterate again: You've said:
> I ended up with this implementation but still if you are super quick after pressing enter to navigate to that url and then open the extension, the tabs query will not have the urls in there.

So _with your PR_, if you open the tab too fast, it still cannot get the URL? Is that correct? (This is just what I doubt)

> I'll open a ticket for Mozilla and ask there?

That's what I'd advise for sure.
.~~Remember also that `onUpdate` won't always help in our case (even if it worked), as when the tab is already open, `onUpdate` is never triggered, as there is no tab update.~~ But, ah, forget taht in this case, we get the URL instantly, of course.

> Merge this one to fix things temporarily and I'll open a ticket for Mozilla and ask there?

Yes.

OogieBoogieInJSON:
umm, where is the merge button ?

rugk:
I have to merge it, of course. :smile: And I'll so right now.

OogieBoogieInJSON:
Ok

rugk:
Oh, I found an issue in the logic: You both throw an error if the maximum retries were exceeded _and_ when the `tab.url` is not there. So either this causes it to always re-try (which it does not seem to do), or it causes it to stop retrying when `tab.url` is `undefined`.

Also the `retryPromise` code is not really clear to me. Should not you e.g. create the `new Promise` only after the original `invokedFunction` promise was rejected?
Maybe try to specify the maximum retry amount in `retryPromise`, too?

OogieBoogieInJSON:
there are no change request left, which one to you refer to ?

rugk:
BTW I've found out why you do not get the `url` when using `browser.tabs.onUpdated`. Actually [the docs say it here](https://developer.mozilla.org/en-US/Add-ons/WebExtensions/API/tabs/Tab):
> Only present if the extension has the "tabs" permission.

That's the catch! This add-on does not have it. (If you add it to the manifest.json it works.) And it does not need it, as it only needs the data of the current active tab, i.e. in production we do not want to require a new permission.

So the fix for the issue may be to actually just do it:
* try first time get active tab (`browser.tabs.query`)
* if it fails (remember it can fail even when the `then` clause is reached), "subscribe" to `browser.tabs.onUpdated`
* (if it is the current tab, by ID, you wanted to get and) the status is "complete" you actually have to re-try to get the data of the active tab via `browser.tabs.query`. This should (always) work then.

What could be tricky is that an error could still happen if you do it as synchronously as I've just described, if the first `browser.tabs.query` is still running and during that time `onUpdated -> completed` is already called.
As such, maybe you need to bind to `onUpdated` earlier and then maybe [race](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Promise/race) these promises against each other? You need to "convert" the [event listener to a Promise though, of course](https://stackoverflow.com/questions/35718645/resolving-a-promise-with-eventlistener). That _may_ be one way… I hope that helps.

---

So as for this code, I think I'll have to say, sorry, but when the approach above works, we do not need the `retryPromise` anymore. I may come back to it, in case we need to retry the Promise.
ANyway, I would not be sure it really catches all cases that may happen in the current way. E.g. [here is async's implementation for the same thing](https://github.com/caolan/async/blob/master/lib/retry.js) and it's, let's say, a little more complicated. Or [this here](https://stackoverflow.com/questions/38213668/promise-retry-design-patterns/44577075#answer-38225011), that looks easier though.

That said,l i guess we may switch to onUpdated now we know why it fails.

OogieBoogieInJSON:
@rugk This is where we first started, the onUpdated listener doesn't have the URL when status becomes "completed". You also said we can do the onUpdate first and then the retry but as I mentioned before, why do both when the retry is more trustworty ?

rugk:
Not retry, you only call it once:
browser.tabs.query -> fails -> wait for `onUpdated` status=completed -> now browser.tabs.query works

Thus, in the end, we only call it twice. At maximum. I've described this in more detail above.

OogieBoogieInJSON:
I did the exact same thing at first. The first browser.tabs.query after onUpdated didn't return the URL :(

rugk:
Can you please (re-)read the beginning of [what I said here](https://github.com/rugk/offline-qr-code/pull/108#issuecomment-405119241)? Or just try it out and add the `"tabs"` permission. It is actually just this permission problem. `onUpdated` in our case will never return the URl, that is why you need to manually call browser.tabs.**query** after you got the status info that it changed.

OogieBoogieInJSON:
There you go, works like a charm!

src/popup/modules/InitQrCode.js
```diff
@@ -86,23 +85,39 @@ export const initiationProcess = Promise.all([qrCreatorInit, userInterfaceInit])
...
-        const tab = tabs[0];
-
-        if (tab && tab.url) {
-            return tab;
-        } else {
-            throw new Error("Url not found.");
-        }
-    }, delay, maxRetries);
+function getCurrentTab() {
+    const queryTabs = () => browser.tabs.query({active: true, currentWindow: true});
```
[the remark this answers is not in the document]
OogieBoogieInJSON: it still returns an array but queryActiveTab should be ok

src/popup/modules/InitQrCode.js
```diff
@@ -86,23 +85,39 @@ export const initiationProcess = Promise.all([qrCreatorInit, userInterfaceInit])
...
+    const bindOnUpdatedListener = (requestedTabId) => {
+        return new Promise((resolve, reject) => {
+            browser.tabs.onUpdated.addListener((tabId, changeInfo, tab) => {
+                if (tab.status === "complete" && tabId === requestedTabId) {
```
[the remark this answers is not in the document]
OogieBoogieInJSON: true that

src/popup/modules/InitQrCode.js
```diff
@@ -86,23 +85,39 @@ export const initiationProcess = Promise.all([qrCreatorInit, userInterfaceInit])
...
+        queryTabs().then((tabs) => {
+            const tab = tabs[0];
+
+            if (tab.url) {
+                resolve(tab);
+
+                return;
+            }
+
+            bindOnUpdatedListener(tab.id).then(resolve).catch(reject);
```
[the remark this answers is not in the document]
OogieBoogieInJSON: That will probably never happen. by the time the browser fires the event with "completed" state the listener was already added.

OogieBoogieInJSON:
I've addressed the last remark mentioned. I won't have time to work on this bug anymore as I'm moving on with something more productive. This already took to much time for the value it actually adds.

rugk:
Okay, I'll take this over then and look whether/how I can adjust the code based on my last review. Thanks for the PR, anyway.

rugk:
So I tinkered around with it a bit and I found out it still does not work when running the `query` immediately after `onComplete` still does not seem to be enough… for whatever reason.

I also tried it with your latest version with the same result: 

grafik

So I'll investigate further and likely make a minimal case to reproduce this issue and report it, as I think now it really has to be a Firefox bug.

rugk:
I'll continue this PR in https://github.com/rugk/offline-qr-code/pull/111.

PR #113, opened 2018-07-25
Use JSON.parse(JSON.stringify(obj)) to log objects

### LINKED ISSUE #28: Properly log objects
Detect whether a thing to be logged is an object and then log it [properly](https://developer.mozilla.org/en-US/docs/Web/API/Console/log#Logging_objects) automatically.

This should be done in the `Logger` of course.

silltho:
Fixes #28

### DIFF
--- a/src/common/modules/Logger.js
+++ b/src/common/modules/Logger.js
@@ -3,6 +3,8 @@ import {ADDON_NAME_SHORT} from "/common/modules/GlobalConsts.js";
 
 import * as AddonSettings from "/common/modules/AddonSettings.js";
 
+import isObject from "/common/modules/lib/lodash/isObject.js";
+
 let debugMode = null;
 
 const MESSAGE_LEVEL_NAME = Object.freeze({
@@ -12,6 +14,22 @@ const MESSAGE_LEVEL_NAME = Object.freeze({
     [MESSAGE_LEVEL.LOADING]: "LOADING",
     [MESSAGE_LEVEL.SUCCESS]: "SUCCESS"
 });
+
+/**
+ * Copy nested objects to ensure a properly output.
+ *
+ * @function
+ * @param  {array} args
+ * @returns {Object}
+ */
+function prepareObjectsForLogging(args) {
+    for(const [index, value] of args.entries()) {
+        if(isObject(value)) {
+            args[index] = JSON.parse(JSON.stringify(value));
+        }
+    }
+    return args;
+}
 
 /**
  * Logs a string to console.
@@ -34,6 +52,8 @@ export function log(...args) {
 
     const messagetype = args[0];
     args[0] = `${ADDON_NAME_SHORT} [${MESSAGE_LEVEL_NAME[messagetype]}]`;
+
+    args = prepareObjectsForLogging(args);
 
     /* eslint-disable no-console */
     switch (messagetype) {
--- a/src/background/modules/BrowserCommunication.js
+++ b/src/background/modules/BrowserCommunication.js
@@ -95,7 +95,7 @@ function saveFileAs(request, sender, sendResponse) {
  * @returns {Promise|null}
  */
 function handleMessages(request, sender, sendResponse) {
-    Logger.logInfo("Got message", JSON.parse(JSON.stringify(request)), "from", JSON.parse(JSON.stringify(sender)));
+    Logger.logInfo("Got message", request, "from", sender);
 
     switch (request.type) {
     case COMMUNICATION_MESSAGE_TYPE.SAVE_FILE_AS:
--- a/src/background/modules/BrowserCommunication.js
+++ b/src/background/modules/BrowserCommunication.js
@@ -95,7 +95,7 @@ function saveFileAs(request, sender, sendResponse) {
  * @returns {Promise|null}
  */
 function handleMessages(request, sender, sendResponse) {
-    Logger.logInfo("Got message", request, "from", sender);
+    Logger.logInfo("Got message", JSON.parse(JSON.stringify(request)), "from", JSON.parse(JSON.stringify(sender)));
 
     switch (request.type) {
     case COMMUNICATION_MESSAGE_TYPE.SAVE_FILE_AS:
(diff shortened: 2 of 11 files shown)


rugk:
Hi @silltho,

silltho:
Hey @rugk, yes I really like your project! :metal:
I'm a pretty new to contributing, so i hope you don't mind that I'm using it to learn. :sweat_smile:

I reworked the PR and hope it is ok now. Pls let me know if i should improve anything.

src/common/modules/Logger.js
```diff
@@ -35,16 +65,18 @@ export function log(...args) {
...
     switch (messagetype) {
-    case MESSAGE_LEVEL.ERROR:
-        console.error(...args);
-        break;
-    case MESSAGE_LEVEL.WARN:
-        console.warn(...args);
-        break;
-    default:
-        console.log(...args);
+        case MESSAGE_LEVEL.ERROR:
```
rugk: Please do not change the indentation.

src/common/modules/Logger.js
```diff
@@ -13,6 +13,36 @@ const MESSAGE_LEVEL_NAME = Object.freeze({
...
 });
 
+/**
+ * Checks if value is a object.
+ *
+ * @function
+ * @param  {*} value
+ * @returns {boolean}
+ */
+function isObject(value){
```
rugk: Lodash is partly included in this project. (and `isObject` is already there) You do not need to copy-paste the code here, just use one good `import` statement. (It's used in other modules here, too.)

src/common/modules/Logger.js
```diff
@@ -13,6 +13,36 @@ const MESSAGE_LEVEL_NAME = Object.freeze({
...
+    const type = typeof value;
+    return value != null && (type === "object" || type === "function");
+}
+
+/**
+ * Copy nested objects to ensure a properly output.
+ *
+ * @function
+ * @param  {...*} args
```
rugk: `{...*}` means can pass infinitely amounts of parameters. That's not true here. You actually do only pass an array.

src/common/modules/Logger.js
```diff
@@ -13,6 +13,36 @@ const MESSAGE_LEVEL_NAME = Object.freeze({
...
+ * @returns {Object}
+ */
+function verifyArgs(args) {
+    for(const key in args) {
+        if (args.hasOwnProperty(key)) {
+            if(isObject(args[key])) {
```
rugk: Well why do you nest that? You could just use "and" (`&&`) to link them. Generally anyway, when you are having a look here: The `Contributing.md` states you should use early return, i.e. check for errors and continue.
Okay, I admit it has to be called "early continue" or so here, but yeah, you may anyway adjust it.

In any case, do whatever you want, but this does not need to be nested.

src/common/modules/Logger.js
```diff
@@ -59,6 +91,8 @@ export function log(...args) {
 export function logError(...args) {
     args.unshift(MESSAGE_LEVEL.ERROR);
 
+    args = verifyArgs(args);
```
rugk: It's enough if you do it once. All these `logXY` also call `log` at the end, so you actually run the args through the whole `verifyArgs` twice. (same below)

src/common/modules/Logger.js
```diff
@@ -13,6 +13,36 @@ const MESSAGE_LEVEL_NAME = Object.freeze({
...
+}
+
+/**
+ * Copy nested objects to ensure a properly output.
+ *
+ * @function
+ * @param  {...*} args
+ * @returns {Object}
+ */
+function verifyArgs(args) {
```
rugk: Better rename it, as it does not really verify any arguments. It more or less just "freezes" the objects(?), or "prepares the objects for logging". Or so? Just come up with a good name. 😄

src/common/modules/Logger.js
```diff
@@ -13,6 +13,36 @@ const MESSAGE_LEVEL_NAME = Object.freeze({
...
+    for(const key in args) {
```
rugk: Remember, `args` is an array. As such you do not need to check `hasOwnProperty` and may better use `for... of`. But I also always confuse it.

Here is the relevant explanation: https://stackoverflow.com/questions/33553944/es6-does-for-of-get-prototype-values-how-to-check-hasownproperty#33554329

src/common/modules/Logger.js
```diff
@@ -13,6 +15,22 @@ const MESSAGE_LEVEL_NAME = Object.freeze({
...
 
+/**
+ * Copy nested objects to ensure a properly output.
+ *
+ * @function
+ * @param  {array} args
+ * @returns {Object}
+ */
+function prepareObjectsForLogging(args) {
+    for(const [index, value] of args.entries()) {
```
rugk: Oh, awesome, you use destructuring here. :smile: 

But this indicates `args` is an object and no array? This can be true and I just completely miss [how the spread syntax worked](https://developer.mozilla.org/docs/Web/JavaScript/Reference/Operators/Spread_syntax). Or, in this case, rather [rest](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Functions/rest_parameters).

silltho: I searched for ways to iterate over an array and also having a index or some other way to exchange the values in it. To me this solution looked the cleanest. But the same could be achieved by using a ```for``` loop.

You are right ```args.entries()``` kind of indicates that ```args``` is an object. So perhaps this solution is still acceptable because ```args``` is kind of an object ? ([The arguments object is an Array-like object](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Functions/arguments))

rugk: No, it's okay, I guess. [entries](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/entries) is a property of an array. I also looked and found only anther way using `.forEach()`, so if you wanted, you may use this, but personally I don't care. This solution here works, too.

BTW `args` is not `arguments` anymore, as we use the rest operator (as linked above), which converts it into an array.

rugk:
So as I tested it I also adjusted some lodash stuff, i.e. using `isPlainObject` instead of `isObject` as I noticed `isPlainObject` is what we actually want. See the lodash doc for that.
Also adjusted some minor other things. See the commit history for that.

PR #115, opened 2018-08-01
Wait for tab URl to be loaded to prevent errors

rugk:
Follow-up of https://github.com/rugk/offline-qr-code/pull/111

It now additionally has a timeout function and a few tweaks to catch the URL as fast as it can and stop re-trying if the "complete" status event has been reached.

I've also finally opened a [Bugzilla issue](https://bugzilla.mozilla.org/show_bug.cgi?id=1479903) with a test WebExtension to reproduce the issue. As such, this code here actually does not work currently, and there is nothing we can do about it.

I consider this mergable.

/cc @OogieBoogieInJSON


rugk:
To be sure, I also tested with dae8ebc and this simple Promise retrying you implemented, @OogieBoogieInJSON, also suffers from the same bug. It seems as soon as the popup is opened, the active tab object somehow get's frozen… :thinking:

PR #134, opened 2018-10-23
Add access key to context menus

### LINKED ISSUE #88: Access keys in context menus
https://discourse.mozilla.org/t/how-to-create-a-menu-with-underscore-as-key-binding/28730

linusaurusrex:
Fixes #88 
Add the underscore as an access key in Firefox 63+.

### DIFF
--- a/src/common/modules/ContextMenu.js
+++ b/src/common/modules/ContextMenu.js
@@ -1,0 +1,26 @@
+import * as Logger from "./Logger.js";
+
+const browserInfo = browser.runtime.getBrowserInfo();
+
+/**
+ * Creates a context menu item
+ * 
+ * @function
+ * @param {string} title 
+ * @param {Object} properties 
+ * @param {function} onCreated 
+ * @returns {Promise}
+ */
+export async function createMenu(title, properties, onCreated) {
+    const info = await browserInfo;
+    const version = parseInt(info.version, 10);
+    if (version > 63) {
+        title += "AccessKey";
+    }
+    
+    properties.title = browser.i18n.getMessage(title);
+    return browser.menus.create(properties, onCreated);
+
+}
+
+Logger.logInfo("ContextMenu module loaded.");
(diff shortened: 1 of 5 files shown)


rugk:
Hi @linusaurusrex,

rugk:
@linusaurusrex 
BTW I want to merge and release this with v1.3 in October still.

linusaurusrex:
1. I used `_` because the link you posted in #88 mentioned an underscore as an access key, but `Q` makes more sense. 
2. Why do you suggest adding a "ContextMenu" module when one already exists?

src/popup/modules/UserInterface.js
```diff
@@ -558,9 +558,9 @@ export function init() {
         }
 
         // create save menu if needed
-        browser.menus.create({
+        browser.runtime.getBrowserInfo().then(({version}) => browser.menus.create({
```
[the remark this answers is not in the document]
rugk: This is still open…

rugk:
Yeah, the ContextMenu module is only for the `background` part. As you also use it in the popup dir, however, it needs to be used by both, so better put it in the `commons` dir.

src/common/modules/ContextMenu.js
```diff
@@ -0,0 +1,26 @@
+import * as Logger from "./Logger.js";
+
+const browserInfo = browser.runtime.getBrowserInfo();
+
+/**
+ * Creates a context menu item
+ * 
+ * @function
```
rugk: I know some old JSDOC still document it like this in this project here, but I'd rather note that this is public, i.e.:
```suggestion
 * @public
```

src/popup/modules/UserInterface.js
```diff
@@ -554,13 +555,12 @@ export function init() {
             // remove menu item if it has been added before
             browser.menus.remove(CONTEXT_MENU_SAVE_IMAGE);
 
-            return;
+            return undefined;
```
rugk: This says not much and you could also have returned nothing AFAIK. However, what about specifically saying that the promise is resolved here:
```suggestion
            return Promise.resolve();
```

PR #135, opened 2018-10-24
Pre-process QR code text

### LINKED ISSUE #41: Remove about:reader from URL
If you open a page in reader mode, it always mangles with the url and uses the I "internally displayed" one.
I think nobody wants these URLs actually, so we could rewrite them to the "original" URL.

Depends on https://github.com/rugk/offline-qr-code/issues/61

### LINKED ISSUE #61: Introduce pre-processing of QR code text
Required for https://github.com/rugk/offline-qr-code/issues/41 and others…

Just some pre-processing of the QR code text – so some things can be changed, which is preferably, possible to disable in the settings.

linusaurusrex:
Fixes #61 and fixes #41

### DIFF
--- a/src/popup/modules/QrCreator.js
+++ b/src/popup/modules/QrCreator.js
@@ -20,8 +20,6 @@ export let qrCreatorInit;
 let initFinished = false;
 let qrCodeLib = null;
 
-let preprocessText = true;
-
 // by default everything has "been changed" (i.e. nothing has been generated yet)
 const changedValues = new Set("text", "color", "size");
 
@@ -86,9 +84,6 @@ export function setSize(size) {
  */
 function preprocess(text) {
     // check for an about:reader URL
-    if (!preprocessText) {
-        return text;
-    }
     const head = text.substring(0, 17);
     const readerUrl = "about:reader?url=";
     if (head === readerUrl) {
@@ -195,8 +190,6 @@ export function init() {
         qrCodeLib.set("qrColor", settings.qrColor);
         qrCodeLib.set("qrBackgroundColor", settings.qrBackgroundColor);
         qrCodeLib.set("qrErrorCorrection", settings.qrErrorCorrection);
-
-        preprocessText = settings.preprocessText;
 
         initFinished = true;
     });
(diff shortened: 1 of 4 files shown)


src/popup/modules/QrCreator.js
```diff
@@ -75,6 +77,27 @@ export function setSize(size) {
...
+ * @function
+ * @param {string} text 
+ * @returns {string}
+ */
+function preprocess(text) {
+    // check for an about:reader URL
+    if (!preprocessText) {
+        return text;
+    }
+    const head = text.substring(0, 17);
```
rugk: Mhh, could not you just use `startsWith` here?

src/popup/modules/QrCreator.js
```diff
@@ -88,8 +111,8 @@ export function setSize(size) {
  * @returns {void}
  */
 export function setText(text) {
-    setTextInternal(text);
-    UserInterface.setQrInputFieldText(text);
+    setTextInternal(preprocess(text));
```
rugk: I know a totally minor "optimisation", but please avoid redoing the same step (pre-processing) again and again. Do it once and use the variable for both.

src/popup/modules/QrCreator.js
```diff
@@ -75,6 +77,27 @@ export function setSize(size) {
...
 
+/**
+ * Pre-processes the text before making the QR code. Changes about:reader URLs to regular URLs.
+ * 
+ * @function
+ * @param {string} text 
+ * @returns {string}
+ */
+function preprocess(text) {
+    // check for an about:reader URL
```
rugk: The comment should be below the if block, should not it? :smile: (otherwise it would have to read "do not check for…"

rugk:
BTW you can change your PR description to "fix[es] #issue", so it [automatically closes](https://help.github.com/articles/closing-issues-via-commit-messages/) the issues when this PR is merged.

rugk:
@linusaurusrex Any answer to my questions/review?

linusaurusrex:
Sorry to respond so slowly. I don't think there needs to be an option to disable removing `about:reader`, but it may be useful to disable other pre-processing depending on what is added. It makes sense then to have add specific options later instead of one pre-processing option.

rugk:
Okay also just spotted one line to change [where to use startsWith and changed it](https://github.com/rugk/offline-qr-code/commit/8bf6e583330849980c585281d86dcfbd2f99a0de) directly when merging.

PR #141, opened 2018-11-04
Adjust new features to Turkish and add changelog

omurturan:
### Changes
1. Add Turkish changelog for 1.3
2. Improve some Turkish translation that did not sound native.
3. Translate "quiet zone" feature
4. Translate access keys

### Known issues
The first letter of the extension is trimmed. It might be because the first letter is "Ç" and there might be strange bug about it. I could not figure it out yet. When I do, I'll do another pull request.

screenshot 2018-11-04 at 17 43 56

### DIFF
--- a/assets/texts/tr/Changelog/1.3.html
+++ b/assets/texts/tr/Changelog/1.3.html
@@ -1,0 +1,14 @@
+<ul>
+<li><strong>Öne çıkan:</strong> Okunurluğu arttırmak için QR kodun çevresine boş bir çerçeve ekleyebilirsin. ("Quite zone" özelliği)</li>
+<li><strong>Yeni:</strong> Testler eklendi.</li>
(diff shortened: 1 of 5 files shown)


rugk:
Thanks!

> The first letter of the extension is trimmed. It might be because the first letter is "Ç" and there might be strange bug about it.

No, I guess this was just me accidentally omitting it when coping it to AMO. (I mean in the downloaded version of the add-on it should be displayed correctly, is not it?)

Talking about AMO: I've added the translations to AMO.

As for the changes in the add-on itself (access keys etc.), I unfortunately cannot really publish them now, because I would have to create a new release – for all users/any language. But they are merged and will be released with the next version. :smile:

rugk:
BTW it's named "quiet zone (for "silent"), not "quite zone". I've fixed that in 66d387b4c0449f345c7ebb91e6fdaece8f5d095b.

PR #156, opened 2019-01-06
upd README.md with Lightning Talk links

Wikinaut:


rugk:
Hmm, no, I see no need /use to put that link there. But thanks for the suggestion, anyway.

Issue #81, opened 2018-05-15
"Create QR code from selection" shown in QR code popup itself


rugk:
## Problem

![grafik](https://user-images.githubusercontent.com/11966684/40078764-e1c9fce4-5885-11e8-9646-fd8567d881b2.png)

Also particularly bad: When you click on that menu item, it does not even ignore it or shows an error, but just closes the popup. This likely happens, because `popup.open` is called, and as it is already opened, this… well… fails.

## Solution

I don't know though, how we could prevent this in an easy way? I doubt binding to "popup closing" is reliable or even possible to e.g. remove the menu item (when popup is shown) and add it again when the popup is hidden.

Binding to the `menuShown` in a similar way [just fails for me](https://bugzilla.mozilla.org/show_bug.cgi?id=1461115) in Firefox 60. I've did that in [this branch](https://github.com/rugk/offline-qr-code/tree/hideOwnContextMenu) (actually just commit https://github.com/rugk/offline-qr-code/commit/0b2ca81e571d51dbb3c7cbc9c3a64ba2a3d9fd6a until now) if someone wants to have a look.

If there any other ideas, I would be glad to hear them.

I also think this is a minor bug, which nobody would notice usually… :wink:

PR #161, opened 2019-01-19
Hide own context menu

rugk:
Just found this old branch that is intended to fix https://github.com/rugk/offline-qr-code/issues/81.

The referenced bug (https://bugzilla.mozilla.org/show_bug.cgi?id=1461115) has been fixed in Firefox 62 apparently, so this just needs some testing.


rugk:
Adjusted and it works now.

PR #168, opened 2019-03-07
Display error/warning for too large input #72

### LINKED ISSUE #72: Display error/warning for too large input
I think the size can be hardcoded… (in a const of course)

https://en.wikipedia.org/wiki/QR_code#Storage

Currently it will just show an empty area…

ongspxm:
close #72 

```
Currently, for data that is too long, `qrCodeLib.genQr()` will return
an error "Data too long". This exception is not caught, and could be
used to display an error message, warning user that the size of data is
too large for the qr code generation.

The error could be displayed using the existing
`CommonMessages.showError()`, an advantage of tying the error messages
to the `oninput` callback is that when the data size shrinks back to
one that can fits into a qrcode, the area will show the actual qr code
for the data again.
```


src/popup/modules/UserInterface.js
```diff
@@ -109,8 +109,15 @@ const refreshQrCode = throttle(() => {
         CommonMessages.hideError();
     }
 
-    QrCreator.setTextInternal(text);
-    QrCreator.generate();
+    try {
+        QrCreator.setTextInternal(text);
+        QrCreator.generate();
+    } catch (e) {
+        if (e === "Data too long") {
```
rugk: Please add some comment explaining where that string comes from etc. (similar to how you do in the PR description, but in short obviously)

src/popup/modules/UserInterface.js
```diff
@@ -109,8 +109,15 @@ const refreshQrCode = throttle(() => {
...
+            CommonMessages.showError("Cannot generate Qr Code, data size too big.");
+            Logger.logError("Data exceeds maximum size:", text.length);
+        }
```
rugk: Please re-throw the error and do _not_ ignore it in case the error is not the one you are catching.

src/popup/modules/UserInterface.js
```diff
@@ -109,8 +109,15 @@ const refreshQrCode = throttle(() => {
...
     }
 
-    QrCreator.setTextInternal(text);
-    QrCreator.generate();
+    try {
+        QrCreator.setTextInternal(text);
+        QrCreator.generate();
+    } catch (e) {
+        if (e === "Data too long") {
+            CommonMessages.showError("Cannot generate Qr Code, data size too big.");
```
rugk: All error messages should be internationalized, so they can be translated/localized into multiple languages. See https://github.com/TinyWebEx/common/blob/master/CONTRIBUTING.md#translations and https://developer.mozilla.org/Add-ons/WebExtensions/Internationalization#Providing_localized_strings_in__locales

Also: _QR_ code… :smile:

ongspxm:
seems like kuajs refuses to throw the code overflow err: https://github.com/lrsjng/kjua/blob/a88369f748926aee8e996e3f720f0f9534c8d86b/src/lib/qrcode.js#L27

removing it from the min

ongspxm:
https://github.com/lrsjng/kjua/blob/a88369f748926aee8e996e3f720f0f9534c8d86b/src/lib/qrcode.js#L27

may need to alter the kjua.min file

ongspxm:
ready for second round of review,

and the kjua thing is kinda hackish, my change is at the end of the min file. search for the only `catch` in the file

src/popup/modules/UserInterface.js
```diff
@@ -109,8 +109,18 @@ const refreshQrCode = throttle(() => {
...
 
-    QrCreator.setTextInternal(text);
-    QrCreator.generate();
+    try {
+        QrCreator.setTextInternal(text);
+        QrCreator.generate();
+    } catch (e) {
+        // Error thrown from qrcodegen & kjua wrapper when code too long
+        if (e === "Data too long") {
+            CommonMessages.showError("errorQrCodeOverflow", true);
```
rugk: Why should it be dismissable? IMHO, it should not.

The error is automatically hidden, when the user inputs valid text later, AFAIK…

src/_locales/en/messages.json
```diff
@@ -80,6 +80,10 @@
     "message": "Permission request failed.",
     "description": "Error shown when a permission request failed. This indicates a technical error and does NOT mean the user declined the permission."
   },
+  "errorQrCodeOverflow": {
+    "message": "Cannot generate QR Code, data size too big.",
```
rugk:
```suggestion
    "message": "Cannot generate QR code, data size too big.",
```

This is how it is written in this project…

rugk:
```suggestion
    "message": "Cannot generate QR Code, the provided text is too much.",
```

also proper sentence and less-technical terms for other users… :wink:

ongspxm:
> So why only for < 40 chars etc.??
I do not understand that change at all…

Haha, there is 40 versions of qrcode, each accepting more characters than the last. What the lib returns is a canvas or image. And when it runs out of the 40 versions or qrcode, it just returns a blank canvas or image.

There isnt a way of determining if its good, because from the outside, the inage is the same. I could try making a pull request to the library repo.

src/popup/modules/UserInterface.js
ongspxm: Means i throw the error regardless? Is that what you meant?

rugk: No, not exactly. I meant the second parameter of the function [`showError`](https://tinywebex.github.io/MessageHandler/module-CommonMessages.html#.showError).

src/popup/modules/QrLib/kjua.js
```diff
@@ -111,8 +111,13 @@ export function set(tag, value) {
  * @returns {HTMLElement}
  */
 export function getQr() {
-    Logger.logInfo("generated new qr kjua code", kjuaOptions);
-    return kjua(kjuaOptions);
+    try {
+        Logger.logInfo("generated new qr kjua code", kjuaOptions);
```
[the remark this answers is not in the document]
rugk: still TODO…

src/popup/modules/UserInterface.js
```diff
@@ -109,8 +109,18 @@ const refreshQrCode = throttle(() => {
...
+            Logger.logError("Data exceeds maximum size:", text.length);
+        } else {
+          throw e;
```
rugk: Indentation error. Do not forget to use EsLint…

src/popup/modules/QrLib/qrerr.js
```diff
@@ -0,0 +1,12 @@
+/**
+ * Errors in QR code generation
+ *
+ * @module QrLib/qrerr
```
rugk: Please just use the long form for the file name. `QrErrors` is not so long one needs to abbreviate it. 😉 

That said, I would have been perfectly fine with including this inside of `QrCreator` or so, but this way it is fine too. 😃

src/popup/modules/QrLib/kjua.js
```diff
@@ -3,12 +3,12 @@
...
  * @module QrLib/kjua
  * @requires /common/modules/Logger
- * @requires qrerr
+ * @requires QrErrors
  */
 /* globals kjua */
 
 import * as Logger from "/common/modules/Logger/Logger.js";
-import * as QrErr from "./qrerr.js";
+import * as QrError from "./QrErrors.js";
```
rugk: Oh, this can easily result in typos. So better either use signular or plural form here... 😄

rugk:
BTW, marking this as blocked by https://github.com/lrsjng/kjua/pull/12

src/popup/modules/UserInterface.js
```diff
@@ -108,8 +110,18 @@ const refreshQrCode = throttle(() => {
...
-    QrCreator.setTextInternal(text);
-    QrCreator.generate();
+    try {
+        QrCreator.setTextInternal(text);
+        QrCreator.generate();
+    } catch (e) {
+        // Error thrown from qrcodegen & kjua wrapper when code too long
+        if (e instanceof QrError.DataOverflowError) {
+            CommonMessages.showError("errorQrCodeOverflow");
+            Logger.logError("Data exceeds maximum size:", text.length);
```
rugk:
```suggestion
            Logger.logError("Maximum size of QR code data exceeded with", text.length, "characters.");
```

The difference is that we do not know whether this is the minimal maximum size, i.e. the maximum size can be 1000 chars, but generation obviously still fails for 2000 chars then. So this log message is just more accurat. 😄

src/popup/modules/QrLib/QrError.js
```diff
@@ -0,0 +1,12 @@
+/**
+ * Errors in QR code generation
+ *
+ * @module QrLib/QrErrors
+ */
+
+export class DataOverflowError extends Error {
+    constructor(...params) {
+        super(...params);
+        this.message = 'data overflow error';
```
rugk: cough... eslint

```suggestion
        this.message = "data overflow error";
```

src/popup/modules/QrLib/kjua.js
```diff
@@ -110,7 +112,13 @@ export function set(tag, value) {
  */
 export function getQr() {
     console.info("generated new qr kjua code", kjuaOptions);
-    return kjua(kjuaOptions);
+
+    try {
+        return kjua(kjuaOptions);
+    } catch (err) {
+        throw err.message.startsWith("code length overflow.")
+          ? new QrError.DataOverflowError() : err;
```
rugk: indentation, see eslint

src/popup/modules/QrLib/qrgen.js
```diff
@@ -99,7 +100,13 @@ export function set(tag, value) {
 export function getQr() {
     console.info("generated new QrGen qr code");
 
-    const qrElem = QRC.encodeText(qrText, qrErrorCorrection);
+    try {
+        const qrElem = QRC.encodeText(qrText, qrErrorCorrection);
```
rugk: and here eslint also catches a logic error that could result in problems/bugs in certain cases 😄

rugk:
Got unblocked as https://github.com/lrsjng/kjua/pull/12 has been merged 😄 

As upstream also updated/changed a little more, can you update the lib [to the new version](https://github.com/lrsjng/kjua/releases/tag/v0.2.0) and test whether it works as expected. (Also as for this change here?)

rugk:
And extra thing: Maybe we can mention somewhere in the source code (or file name?), which (in this case) kjua version we are using?

rugk:
It's version 0.2.0, not 2.0... 😃

rugk:
* [x] verified that the included kjua lib is [the original one](https://github.com/lrsjng/kjua/blob/32565e6eeee95014c11c753b492c6ae69d4bcb07/dist/kjua.min.js)

src/_locales/en/messages.json
```diff
@@ -80,6 +80,10 @@
     "message": "Permission request failed.",
     "description": "Error shown when a permission request failed. This indicates a technical error and does NOT mean the user declined the permission."
   },
+  "errorQrCodeOverflow": {
+    "message": "Cannot generate QR code, the provided text is too much.",
```
rugk: I know, I am correcting myself, but as per https://design.firefox.com/photon/copy/writing-for-users.html#be-human I'd rather choose something that sounds better than trying to be 100% technically correct.

I.e.:
```suggestion
    "message": "Cannot generate QR code, the provided text is too long.",
```

("much text" sounds awkward IMHO)

src/popup/modules/QrLib/QrError.js
```diff
@@ -0,0 +1,12 @@
+/**
+ * Errors in QR code generation
+ *
+ * @module QrLib/QrErrors
+ */
+
+export class DataOverflowError extends Error {
+    constructor(...params) {
+        super(...params);
+        this.message = "data overflow error";
```
rugk: So when we are at it: Can you change it to only _optionally_ use this as a default message? If one manually does `new DataOverflowError("custom message")` that should still be honored and not being dismissed.

rugk: Also the message may be more human-readable:
```suggestion
        this.message = "The QR code was given too much data.";
```

ongspxm:
@rugk thanks for the fast review, ready again

src/popup/modules/QrLib/QrError.js
```diff
@@ -0,0 +1,13 @@
+/**
+ * Errors in QR code generation
+ *
+ * @module QrLib/QrErrors
+ */
+
+export class DataOverflowError extends Error {
+    constructor(...rparams) {
```
rugk:
```suggestion
    constructor(message, ...rparams) {
```

Better start like this...

src/popup/modules/QrLib/QrError.js
```diff
@@ -0,0 +1,13 @@
...
+        let params = rparams.length>0
```
rugk: You do not need to make it so compact. Better be a little more elaborate, but make actual something easily readable.

Better use a real if and only overwrite the first parameter, if given. (See suggestion above.)

src/popup/modules/UserInterface.js
```diff
@@ -108,8 +109,18 @@ const refreshQrCode = throttle(() => {
         CommonMessages.hideError();
     }
 
-    QrCreator.setTextInternal(text);
-    QrCreator.generate();
+    try {
+        QrCreator.setTextInternal(text);
+        QrCreator.generate();
+    } catch (e) {
+        // Error thrown from qrcodegen & kjua wrapper when code too long
```
rugk: Which code? 😄 

```suggestion
        // error thrown from qrcodegen & kjua wrapper when text input is too long
```

src/_locales/en/messages.json
rugk: Any comment here? You did not include this change...

src/popup/modules/QrLib/QrError.js
```diff
@@ -0,0 +1,13 @@
+/**
+ * Errors in QR code generation
+ *
+ * @module QrLib/QrErrors
```
rugk:
```suggestion
 * @module QrLib/QrError
```

PR #97, opened 2018-06-03
Use browser-style for options

rugk:
Fixes https://github.com/rugk/offline-qr-code/issues/44

grafik

blocked because of:
* https://bugzilla.mozilla.org/show_bug.cgi?id=1466420

also it would be nice to have these solved first:
* https://bugzilla.mozilla.org/show_bug.cgi?id=1466419
* https://bugzilla.mozilla.org/show_bug.cgi?id=1466418

Issue #44, opened 2018-04-23
Adjust style of input elements to Firefox ones


rugk:
Again as `about:preferences` does it, or, as [the Photon styling guide](https://design.firefox.com/photon/components/checkboxes.html) states how it should be done.

This may require quite "much" CSS, but it can partically be taken from the linked photon site.

It should be applied to: checkboxes, radio elements, etc.

Note that it my look strange when it is applied to one element, but not all elements. So it needs to be done for all of them.

May be done in conjunction with https://github.com/rugk/offline-qr-code/issues/43.

rugk:
Note that as of Firefox 60 the photon design has not really been applied to the elements on the options page or so… So maybge be wait or use the "old" design. (I assume the design will be adjusted, later.)

rugk:
Uuuuh… wait… I cannot read the doc.

https://developer.mozilla.org/en-US/Add-ons/WebExtensions/user_interface/Browser_styles says the class `browser-style` still needs to be applied to some elements to work…

![image](https://user-images.githubusercontent.com/11966684/39678086-653dbf7c-5186-11e8-813c-0cbaeef4c6ce.png)

So this is way easier than I thought. :smile: 
And there they actually follow the Photon design (for checkboxes e.g.) – in contrast to the built-in settings page in Firefox. :laughing:

rugk:
It may still require some tweaking so you cannot just stupidly apply the class. E.g. this one looks crazy when applied to my select element:
![image](https://user-images.githubusercontent.com/11966684/39678097-96589bae-5186-11e8-85aa-edfc1901d9d9.png)

PR #170, opened 2019-03-14
Adjust and group options style with headers like Firefox #43

### LINKED ISSUE #43: Adjust and group options style with headers like Firefox
Using `about:preferences` as a template, the aim should be to have something simlar from a grouping perspective.

This also means to drop `fieldset`s (the border around the radio options as a "radio group").

See https://design.firefox.com/photon/visuals/grid.html

ongspxm:
closes #43 

```
Currently the options page is not inline with the style guide used by 
firefox's setting pages (e.g. `about:preferences`).

Let's update the option page to use similar styles and at the same 
time, add some grouping to the options to enhance the page.
```

### DIFF
--- a/src/_locales/en/messages.json
+++ b/src/_locales/en/messages.json
@@ -197,7 +197,7 @@
     "description": "The message, which appears, when settings are pre-defined (as managed options) by administrators."
   },
   "titleGeneral": {
-    "message": "General Settings",
+    "message": "Add On",
     "description": "The title for the general settings group."
   },
   "titleQrCodeSetting": {
@@ -221,10 +221,6 @@
     "description": "The button to delete all current settings and load the defaults, shown in the add-on settings."
   },
 
-  "subtitlePopupIconColored": {
-    "message": "Colored icon",
-    "description": "This is a sub heading in the add-on settings."
-  },
   "optionPopupIconColoredDescr": {
     "message": "Shows a colored icon instead of the black/white icon in the toolbar.",
     "description": "This is an option shown in the add-on settings. It describes the optionPopupIconColored setting in more details."
--- a/src/_locales/en/messages.json
+++ b/src/_locales/en/messages.json
@@ -201,11 +201,11 @@
     "description": "The title for the general settings group."
   },
   "titleQrCodeSetting": {
-    "message": "QR Code",
+    "message": "QR code",
     "description": "The title for the qr code settings group."
   },
   "titleAddonBehavior": {
-    "message": "Add-on Behavior",
+    "message": "Add-on behavior",
     "description": "The title for the addon behavior settings group."
   },
   "optionIsDisabledBecauseManaged": {
@@ -277,7 +277,7 @@
   },
 
   "subtitleMisc": {
-    "message": "Miscellaneous Settings",
+    "message": "Miscellaneous settings",
     "description": "The heading/grouping title shown for the miscellaneous setting of the QR code."
   },
   "optionQrQuietZone": {
--- a/src/_locales/en/messages.json
+++ b/src/_locales/en/messages.json
@@ -221,16 +221,16 @@
     "description": "The button to delete all current settings and load the defaults, shown in the add-on settings."
   },
 
-  "optionPopupIconColored": {
+  "subtitlePopupIconColored": {
     "message": "Colored icon",
-    "description": "This is an option shown in the add-on settings."
+    "description": "This is a sub heading in the add-on settings."
   },
   "optionPopupIconColoredDescr": {
     "message": "Shows a colored icon instead of the black/white icon in the toolbar.",
     "description": "This is an option shown in the add-on settings. It describes the optionPopupIconColored setting in more details."
   },
 
-  "optionQrCodeType": {
+  "subtitleQrCodeType": {
     "message": "QR code type",
     "description": "The heading/grouping title shown for the radio buttons for the type setting of the QR code."
   },
@@ -251,13 +251,9 @@
     "description": "Helper text for canvas QR code type option."
   },
 
-  "optionQrCodeSize": {
+  "subtitleQrCodeSize": {
     "message": "QR code size",
     "description": "The heading/grouping title shown for the radio buttons for the size setting of the QR code."
-  },
-  "optionMisc": {
-    "message": "Miscellaneous Settings",
-    "description": "The heading/grouping title shown for the miscellaneous setting of the QR code."
   },
   "optionQrCodeSizeFixed": {
     "message": "Fixed size:",
@@ -280,6 +276,10 @@
     "description": "This is an option shown in the add-on settings. It is one option for selecting the QR code size."
   },
 
+  "subtitleMisc": {
+    "message": "Miscellaneous Settings",
+    "description": "The heading/grouping title shown for the miscellaneous setting of the QR code."
+  },
   "optionQrQuietZone": {
     "message": "QR quiet zone:",
     "description": "This is an option shown in the add-on settings. It's about the area around the QR code."
--- a/src/_locales/en/messages.json
+++ b/src/_locales/en/messages.json
@@ -221,6 +221,10 @@
     "description": "The button to delete all current settings and load the defaults, shown in the add-on settings."
   },
 
+  "optionPopupIconColored": {
+    "message": "Colored icon",
+      "description": "This is an option shown in the add-on settings."
+  },
   "optionPopupIconColoredDescr": {
     "message": "Shows a colored icon instead of the black/white icon in the toolbar.",
     "description": "This is an option shown in the add-on settings. It describes the optionPopupIconColored setting in more details."
--- a/src/_locales/en/messages.json
+++ b/src/_locales/en/messages.json
@@ -196,6 +196,18 @@
     "message": "Some settings are managed by your system administrator and cannot be changed.",
     "description": "The message, which appears, when settings are pre-defined (as managed options) by administrators."
   },
+  "titleGeneral": {
+    "message": "General Settings",
+    "description": "The title for the general settings group."
+  },
+  "titleQrCodeSetting": {
+    "message": "QR Code Settings",
+    "description": "The title for the qr code settings group."
+  },
+  "titleAddonBehavior": {
+    "message": "Addon Behavior",
+    "description": "The title for the addon behavior settings group."
+  },
   "optionIsDisabledBecauseManaged": {
     "message": "This option is disabled, because it has been configured by your system administrator.",
     "description": "The title (tooltip) shown, when hovering over a disabled, managed option."
@@ -242,6 +254,10 @@
   "optionQrCodeSize": {
     "message": "QR code size",
     "description": "The heading/grouping title shown for the radio buttons for the size setting of the QR code."
+  },
+  "optionMisc": {
+    "message": "Miscellaneous Settings",
+    "description": "The heading/grouping title shown for the miscellaneous setting of the QR code."
   },
   "optionQrCodeSizeFixed": {
     "message": "Fixed size:",
--- a/src/options/options.css
+++ b/src/options/options.css
@@ -19,10 +19,6 @@ body.mobile .mobile-incompatible {
   opacity: 0.4;
 }
 
-form {
-  margin: 16px 32px;
-}
-
 section {
   margin-bottom: 32px;
 }
@@ -46,6 +42,7 @@ h1 {
 
 form section:first-child h1 {
   border-top: none;
+  margin-top: 16px;
 }
 
 h2{
--- a/src/options/options.css
+++ b/src/options/options.css
@@ -1,7 +1,6 @@
 body {
-  color: #333;
-  font-family: sans-serif;
-  font-size: 14px;
+  /* body20 */
+  font-size: 15px;
 }
 
 /* on (small) mobile displays */
@@ -28,9 +27,10 @@ h1 {
   margin: 0;
   padding: 16px 0;
 
+  /* title30 */
+  font-size: 22px;
   border-top: 1px solid rgba(12, 12, 13, 0.15);
   font-weight: 300;
-  font-size: 1.46em;
 }
 
 h1:first-child {
@@ -38,9 +38,10 @@ h1:first-child {
 }
 
 h2{
+  /* title20 */
+  font-size: 17px;
   margin-bottom: 4px;
   font-weight: 600;
-  font-size: 1.14em;
 }
 
 ul {
@@ -63,6 +64,7 @@ li {
 }
 
 .item .line * {
+  /* https://design.firefox.com/photon/visuals/grid.html#spacing */
   margin: 0 8px 0 0;
 }
 
@@ -71,6 +73,8 @@ li {
 }
 
 .helper-text {
+  /* caption20 */
+  font-size: 13px;
   color: var(--grey-50);
 }
 
--- a/src/options/options.css
+++ b/src/options/options.css
@@ -1,3 +1,7 @@
+:root {
+  --gray: rgba(12, 12, 13, 0.15);
+}
+
 body {
   /* body20 */
   font-size: 15px;
@@ -27,7 +31,7 @@ section {
 hr {
   border: 0;
   height: 1px;
-  background-color: rgba(12, 12, 13, 0.15);
+  background-color: var(--gray);
 }
 
 h1 {
@@ -38,7 +42,7 @@ h1 {
   font-size: 22px;
   font-weight: 300;
 
-  border-top: 1px solid rgba(12, 12, 13, 0.15);
+  border-top: 1px solid var(--gray);
 }
 
 form section:first-child h1 {
--- a/src/options/options.css
+++ b/src/options/options.css
@@ -37,14 +37,21 @@ h1:first-child {
   border-top: none;
 }
 
-section {
-  margin-bottom: 32px;
-}
-
 h2{
   margin-bottom: 4px;
   font-weight: 600;
   font-size: 1.14em;
+}
+
+ul {
+  padding: 0;
+  margin-bottom: 32px;
+}
+
+li {
+  list-style-type: none;
+  margin-top: 10px;
+  padding: 0px;
 }
 
 .line {
--- a/src/options/options.css
+++ b/src/options/options.css
@@ -23,9 +23,13 @@ form {
   margin: 16px 32px;
 }
 
+section {
+  margin-bottom: 32px;
+}
+
 h1 {
-  margin: 0;
-  padding: 16px 0;
+  margin: 32px 0 0 0;
+  padding: 16px 0 0 0;
 
   /* title30 */
   font-size: 22px;
@@ -41,20 +45,20 @@ h1.first {
 h2{
   /* title20 */
   font-size: 17px;
-  margin-bottom: 4px;
   font-weight: 600;
+  margin-top: 32px;
 }
 
 ul {
+  margin: 0;
   padding: 0;
-  margin-bottom: 32px;
 }
 
 li {
   list-style-type: none;
 
   padding: 2px 0;
-  margin: 10px 0 4px 0;
+  margin: 0 0 4px 0;
 
   display: flex;
   align-items: center;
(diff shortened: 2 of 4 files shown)


src/options/options.html
```diff
@@ -59,110 +59,156 @@
...
-					<label data-i18n="__MSG_optionDebugMode__" for="debugMode">Enable debug mode</label>
-					<span data-i18n="__MSG_optionDebugModeDescr__" class="helper-text">This is only useful to get more detailed logging output in the console for reporting bugs, etc.</span>
-				</li>
-				<li>
-					<button data-i18n="__MSG_optionsResetButton__" type="button" id="resetButton">Reset all settings to defaults</button>
-				</li>
-			</ul>
+					</div>
+			</section>
+			<h1>TODO: Addon behavior</h1>
```
rugk: It's possibly better to not make these bigger headers h1, but h2.

rugk:
Note the details may actually depend on https://github.com/rugk/offline-qr-code/pull/97, respectivly merging https://github.com/rugk/offline-qr-code/pull/97 would increase familarily a lot, too, but as you see this is blocked for several reasons.

src/options/options.html
ongspxm: my rationale is to reduce the amount of "defaults removing behavior", ul and li have those listing properties. i could add that back in

ongspxm:
> Note the details may actually depend on #97, respectivly merging #97 would increase familarily a lot, too, but as you see this is blocked for several reasons.

okay, i will go and take a look

rugk:
No, that was not how I meant it. That was purely informational, and you can ignore #97. Actually, solving #97 is a totally seperate task, so just keep this PR here, as it is. #97 can be merged after/seperately or none at all afterwards.

src/options/options.html
```diff
@@ -1,214 +1,214 @@
 <!DOCTYPE html>
 <html lang="en">
-	<head>
-		<meta charset="utf-8">
-		<meta name="viewport" content="width=device-width, initial-scale=1.0">
-		<link rel="stylesheet" href="../common/common.css">
-		<link rel="stylesheet" href="options.css">
+  <head>
```
rugk: Please do not change indentation and use `editorconfig`, which [requires tabs for HTML](https://github.com/rugk/offline-qr-code/blob/master/.editorconfig#L21).

ongspxm:
> No, that was not how I meant it. That was purely informational, and you can ignore

Do you prefer to use `browser-style`?

ongspxm:
> No, that was not how I meant it. That was purely informational, and you can ignore

Do you prefer to use `browser-style`?

rugk:
No, as said, this is the task of https://github.com/rugk/offline-qr-code/pull/97 and https://github.com/rugk/offline-qr-code/issues/44, so do not add this. Reasons are stated in https://github.com/rugk/offline-qr-code/pull/97.

src/_locales/en/messages.json
```diff
@@ -243,6 +255,10 @@
     "message": "QR code size",
     "description": "The heading/grouping title shown for the radio buttons for the size setting of the QR code."
   },
+  "optionMisc": {
```
rugk: Well… for consistency you should then also name this "title…`, should not you?

src/options/options.css
```diff
@@ -20,51 +20,51 @@ body.mobile .mobile-incompatible {
...
+  margin-bottom: 4px;
+
+  display: flex;
+  align-items: center;
 }
 
-.helper-text {
-  display: block;
-  color: var(--grey-50);
+.item .line * {
```
rugk: Is this `*` selector really needed? :thinking:

ongspxm:
I used title to refer to the h1 text, this is one of the subtitles, so i used the current naming convention omstead

src/options/options.css
ongspxm: This is to add the 8px padding to all first class elements in the row, as per the firefox design.

src/_locales/en/messages.json
rugk: > I used title to refer to the h1 text, this is one of the subtitles, so i used the current naming convention omstead

So let's keep the discussion here in this line.
Okay, fair enough. However, then maybe name it `subtitle` or so?

I mean we always have `description`, where we can explain it all in detail...

src/options/options.css
```diff
@@ -1,7 +1,7 @@
 body {
-  /* the style the other options in the settings pane use */
-  font-size: 1.11em;
   color: #333;
+  font-family: sans-serif;
```
rugk: Okay, on Windows this results in `Arial` and `Tahoma` instead of `Segoe UI` and `Tahoma`.

## Arial (with `sans-serif`)

grafik

## Segoe UI (without `sans-serif`)

grafik

Note that we use a basic `browser-style` for [the optiosn page](https://developer.mozilla.org/en-US/docs/Mozilla/Add-ons/WebExtensions/user_interface/Options_pages) (see at the bottom), so AFAIK the best font should be provided by Firefox itself. See also https://stackoverflow.com/questions/39830037/web-extension-how-do-i-use-browser-style-true

ANd IMHO the "Segoe UI" version also looks closer to the original Firefox settings page than Arial.

src/options/options.css
```diff
@@ -20,51 +20,51 @@ body.mobile .mobile-incompatible {
...
-.condensed-list {
-  margin-left: 0px;
+
+h1 {
+  margin: 0;
+  padding: 16px 0;
+
+  border-top: 1px solid rgba(12, 12, 13, 0.15);
+  font-weight: 300;
+  font-size: 1.46em;
```
rugk: What makes you choose this font-size? [The photon guide](https://design.firefox.com/photon/visuals/typography.html#scale) lists the [title](https://design.firefox.com/photon/visuals/typography.html#title) as "Title 30", which seems to be `22px`.

src/options/options.css
```diff
@@ -20,51 +20,51 @@ body.mobile .mobile-incompatible {
...
-  width: auto;
-  min-width: 60vw;
+section {
+  margin-bottom: 32px;
+}
 
+h2{
   margin-bottom: 4px;
+  font-weight: 600;
+  font-size: 1.14em;
```
rugk: same font-size question here. This is suppsoed to be "Title 20" = 17px

src/options/options.css
rugk: Okay, another image:

grafik

Looks better this way. 😃 

That said, do you have a link where the Photon guide states this?

rugk: Also consider to add references/some explanation to the CSS source, if needed.

ongspxm: limiting the 8px margin to only elements used, is kinda like *

src/options/options.css
ongspxm: oh, i dug into the actual css. okay, i will follow this instead

src/options/options.css
ongspxm: the `browser-style` uses the browser's default font setting instead. So for example, if my default font is a mono-font, then it will use the mono-font. This is different from the font `about:preference` uses.

So on my own setting, i get a mono font for the option page, but a sans-serif font for about:preference

ongspxm: but either ways, i can remove this, so it will use the user's default font settings.

src/options/options.css
rugk: Yes, it may be actualyl result in the same value in calculations. Possibly... 😉 But better to use the documented values. 😄

src/options/options.html
```diff
@@ -59,157 +59,155 @@
 			</div>
 		</div>
 		<form>
-			<h1 data-i18n="__MSG_titleGeneral__">General</h1>
-			<ul class="mobile-incompatible">
+			<h1 class="first" data-i18n="__MSG_titleGeneral__">General</h1>
```
rugk: Headings should be _inside_ of a section. See https://developer.mozilla.org/en-US/docs/Web/HTML/Element/section:
> Each <section> should be identified, typically by including a heading (<h1>-<h6> element) as a child of the <section> element.

ongspxm:
@rugk you are really quick with this :+1: thanks so much. ready again

rugk:
BTW, if I understand the GitHub GUI correctly, you should also have a button somewhere here, where you can click to "request a new review". 😃

ongspxm:
> BTW, if I understand the GitHub GUI correctly, you should also have a button somewhere here, where you can click to "request a new review". smiley

i cannot seem to request for reviewers, seems like some permission thing, not too sure how this works. are you talking about the reviewers box in the top right hand corner?

rugk:
I thought of this at the bottom:
grafik

Yes, the reviewer box at the top right likely does not work.

src/options/options.css
```diff
@@ -20,51 +19,64 @@ body.mobile .mobile-incompatible {
...
+
+section {
+  margin-bottom: 32px;
 }
-/* in a fieldset e.g. we use a condensed version */
-.condensed-list {
-  margin-left: 0px;
+
+h1 {
+  margin: 32px 0 0 0;
```
rugk: So why not just...?
```suggestion
  margin-top: 32px;
```
(untested)

(same for many others)

ongspxm:
The default margins of the headers, ul and li was the reason so the big
gaps previously.

ongspxm:
image

haha, mine looks different, i think i no rights

ongspxm:
screenshot-1553096054_332x375

at the lastest commit, i run into this error, could you shed some light on how to proceed? how do we set the default value? the original code doesnt seem to have a default value set.

rugk:
Okay, great of GitHub I can re-request a review from myself, but you cannot. 😆

rugk:
The default values are saved in a JS file in the `common/data` dir, see https://github.com/TinyWebEx/AddonSettings#default-values.

The problem is, you set triggers for each single radio value to save it as an option. This won't work, as it now tries to load an option for . So better revert that... 😉 

Actually, loading all worked, only saving failed. So don't also break loading the option. 😉 

The issue is, the "Qr code size" setting is actually grouped into a JSON obejct before being saved. See https://github.com/TinyWebEx/AutomaticSettings#option-groups and https://github.com/rugk/offline-qr-code/blob/master/src/common/modules/data/DefaultSettings.js#L17-L20

ongspxm:
finally got it figured out, the docs helped a lot, thanks

src/options/options.html
```diff
@@ -59,110 +59,153 @@
...
-					<span>
-						<span id="qrQuietZoneStatus">(1 square)</span>
-					</span>
-					<span class="helper-text">
+			<section class="mobile-incompatible">
+				<h1 class="first" data-i18n="__MSG_titleGeneral__">Add On</h1>
+				<ul>
+					<li>
+						<input class="setting save-on-change" type="checkbox" id="popupIconColored" name="popupIconColored">
+						<label data-i18n="__MSG_optionPopupIconColoredDescr__" for="popupIconColored" >Shows a colored icon instead of the black/white icon in the toolbar.</label>
```
rugk: BTW, as for this option, I think it sounds strange now (the label is kinda elaborate). Remember, before it was a label + a helper text:
grafik

Maybe we can restore it like this?

src/_locales/en/messages.json
```diff
@@ -196,6 +196,18 @@
     "message": "Some settings are managed by your system administrator and cannot be changed.",
     "description": "The message, which appears, when settings are pre-defined (as managed options) by administrators."
   },
+  "titleGeneral": {
+    "message": "Add On",
+    "description": "The title for the general settings group."
+  },
+  "titleQrCodeSetting": {
+    "message": "QR Code Settings",
```
rugk: Possibly just like this?

```suggestion
    "message": "QR Code",
```

I mean, all options there are settings…

But if you have a different idea, feel free to comment.
Also, obviously, if you e.g. would re-order the options, that is totally possible. (e.g. monospace font is also only shown in the QR code popup, but mhh… I kinda like it this way and would also have no idea how to improve it.)

src/_locales/en/messages.json
```diff
@@ -196,6 +196,18 @@
...
+    "description": "The title for the qr code settings group."
+  },
+  "titleAddonBehavior": {
+    "message": "Addon Behavior",
```
rugk:
```suggestion
    "message": "Add-on Behavior",
```

https://design.firefox.com/photon/copy/word-list.html#add-on strikes again :wink:

src/_locales/en/messages.json
```diff
@@ -196,6 +196,18 @@
     "message": "Some settings are managed by your system administrator and cannot be changed.",
     "description": "The message, which appears, when settings are pre-defined (as managed options) by administrators."
   },
+  "titleGeneral": {
+    "message": "Add On",
```
rugk:
```suggestion
    "message": "Add-on",
```
https://design.firefox.com/photon/copy/word-list.html#add-on

src/options/options.css
```diff
@@ -20,74 +19,73 @@ body.mobile .mobile-incompatible {
...
+/* when a link is used in a helper text, add margin */
+.helper-text > a {
+  margin-left: 4px;
+  overflow-wrap: none;
 }
 
-#qrCodeSizeFixedValue {
-  width: 4.5em;
+#qrQuietZone, label {
+  cursor: pointer;
```
rugk: Why pointer style? (also for the checkbox labels, as it seems)

In `about:preferences` there is nowhere such a pointer style.

src/options/options.css
```diff
@@ -20,74 +19,73 @@ body.mobile .mobile-incompatible {
...
 
-/* special options values */
+li.indent {
+  margin-inline-start: 28px;
+}
 
-#popupIconColor {
+li.helper-text {
+  /* caption20 */
+  font-size: 13px;
```
rugk: I have the feeling that this is a little too small, or at least smaller than in the Firefox settings page. Also in the screenshot on https://design.firefox.com/photon/patterns/helpers.html#necessary, it looks bigger. (maybe just like usual text?)

rugk:
Another note: This needs testing on Android, as there the general font-size of the settings is different and [there I adjust the design](https://github.com/TinyWebEx/AutomaticSettings/blob/276f19c2862cd6ec3d91ff64332f2742da99f616/MobileOptions.js#L44).

src/options/options.css
ongspxm: i felt it would be clear make it clear that its an input option, that clicking it does something. but i could remove it. do you want it removed?

ongspxm:
> Another note: This needs testing on Android, as there the general font-size of the settings is different and there I adjust the design.

For that i need to build the xpi file right?

rugk:
> For that i need to build the xpi file right? is it just running the build script?

Yep, you can do. (Note you must run it from the repo dir, i.e. don't cd into `scripts`.)
Or you just ZIP the whole dir (`manifest.json` should be in the root), then you may only have some more, unnecessary files. You can follow the official docs on MDN then.

src/options/options.css
```diff
@@ -20,74 +19,69 @@ body.mobile .mobile-incompatible {
...
+  font-weight: 300;
 
-  margin-bottom: 4px;
+  border-top: 1px solid rgba(12, 12, 13, 0.15);
 }
 
-/* when a setting has a label *before* it, add some margin */
-label + .setting {
-  margin-left: 8px;
+h1.first {
```
rugk: Hmm, that markup in the HTML is not so nice…

Can't we just use some pseudo-selectors of CSS in a clever way to achieve this?
Kinda like: https://stackoverflow.com/questions/2717480/css-selector-for-first-element-with-class

I'll guess find a good way… :wink:

src/options/options.css
```diff
@@ -20,74 +19,69 @@ body.mobile .mobile-incompatible {
...
-ul {
-  margin: 0px;
-  padding: 0px;
-}
-li {
-  list-style-type: none;
-  margin-top: 10px;
-  padding: 0px;
+form {
+  margin: 16px 32px;
```
rugk: Where did you take this numbers from? Are these somewhat official?

Because if I had to choose a margin/padding for that here, I'd try to align the options with how the other Firefox elements are inset at the top of it?
Is this possible in some way? If so, it may look quite nice as it aligns better into the general UI then…

align


(And BTW, I predict on Android this can cause issues :wink:)

ongspxm: screenshot-1553834447_667x821

this is how it looks without margin, does it looks a little squeezy?

ongspxm:
does it look alright on android?

test

src/options/options.css
rugk: Do _you_ think so?

src/options/options.css
```diff
@@ -38,7 +44,7 @@ h1 {
   border-top: 1px solid rgba(12, 12, 13, 0.15);
 }
 
-h1.first {
+form section:first-child h1 {
```
rugk: Yeah, great solution! 👍

src/options/options.css
```diff
@@ -19,14 +19,16 @@ body.mobile .mobile-incompatible {
...
 
-form {
-  margin: 16px 32px;
-}
-
 section {
   margin-bottom: 32px;
 }
 
+hr {
```
rugk: Yeah...

Just one thing here:
As you have now copy & pasted the code from teh h1 tag, you have code duplication. This is not good.

It may be better if this can be changed...
Okay, I see the CSS tags are different, but possibly... (just would be nice, I totally see there is some difference here.)

src/options/options.html
```diff
@@ -207,6 +207,7 @@ <h1 data-i18n="__MSG_titleAddonBehavior__">Addon behavior</h1>
 			</section>
 
 			<section>
+				<hr />
```
rugk: Okay, for some reason this does seem to shrink the size of the button, does not it?

It is so "flat" now... (in height)
grafik

Maybe this can be changed?

Also, generally, do you like the seperation line there, or do you think it is not really needed and just confuses users?

src/options/options.css
ongspxm: the sides at least

src/options/options.html
ongspxm: do you think its better to use a br instead? just put more padding to show that its different?

src/options/options.css
rugk: Okay fine. Go with the way you like it best… (maybe even a compromise, i.e. not the "original" 32px or what it was)

src/options/options.html
rugk: I honestly am not sure. Maybe try and see whether it looks better…

Note that in any case you can keep the CSS for the `hr`… Maybe it can be used later anyway… :wink: 

Note I am also thinking about https://github.com/rugk/offline-qr-code/issues/173 and there we could have some more content at the bottom, which may look totally different then…

So I am open to suggestions/ideas here… :smiley:

ongspxm: i changed it to 2 `br` looks not too bad

src/options/options.css
```diff
@@ -1,6 +1,11 @@
+:root {
+  --gray: rgba(12, 12, 13, 0.15);
```
rugk: Is this an official Photon variable?

If so, then you could also [include it here](https://github.com/rugk/offline-qr-code/blob/master/src/common/variables.css). (Or if it is useful in other contexts.)

Note you'd better leave it as it is, if it is no official variable.

src/options/options.css
```diff
@@ -1,7 +1,11 @@
...
```
rugk: Or, different idea, maybe rename it for the actual use case?

```suggestion
  --gray-separator: rgba(12, 12, 13, 0.15);
```

src/options/options.css
```diff
@@ -20,74 +24,77 @@ body.mobile .mobile-incompatible {
...
-  padding: 0px;
-}
-/* in a fieldset e.g. we use a condensed version */
-.condensed-list {
-  margin-left: 0px;
+
+hr {
+  border: 0;
+  height: 1px;
+  background-color: var(--gray);
```
ongspxm:
```suggestion
  background-color: var(--gray-separator);
```

src/options/options.css
```diff
@@ -20,74 +24,77 @@ body.mobile .mobile-incompatible {
...
-  /* a big width, but not too big */
-  display: inline-block;
-  width: auto;
-  min-width: 60vw;
+  /* title30 */
+  font-size: 22px;
+  font-weight: 300;
 
-  margin-bottom: 4px;
+  border-top: 1px solid var(--gray);
```
ongspxm:
```suggestion
  border-top: 1px solid var(--gray-separator);
```

PR #184, opened 2019-04-23
AMO text zh-CN translate

Cye3s:


assets/texts/zh-CN/Changelog/1.1.html
```diff
@@ -0,0 +1,12 @@
+<ul>
```
rugk: You don't need to copy the changelog entries if you are not going to translate them. :smiley:

assets/texts/zh-CN/amoScreenshots.csv
```diff
@@ -0,0 +1,8 @@
+qrBig.png; "QR codes are resizable!";
```
rugk: These descriptions do also appear on AMO, so it would make quite sense to translate them, at least. :smile:

PR #199, opened 2019-06-02
Issue #196 - Remove old FF workaround setTimeout

### LINKED ISSUE #196: Test whether auto-select retry is still needed with Firefox >= 60
I've found these comments:

https://github.com/rugk/offline-qr-code/blob/e900c8878f86f2b3bea5558b686097b766fa9cba/src/popup/modules/UserInterface.js#L191

https://github.com/rugk/offline-qr-code/blob/e900c8878f86f2b3bea5558b686097b766fa9cba/src/popup/modules/UserInterface.js#L162

The code is responsible for selecting all the text and scrolling to the top when you open the QR code popup.

The comments indicate this is a workaround only for Firefox <= 60, but unfortunately no bug is linked or so.

However, nowadays we require Firefox 60 as a minimum version, anyway.

## TODO

* So we need to test – if we remove these workarounds –, does it still work in Firefox >= 60?
* If so, remove it from the code. (and clean up related code, if needed)

cornu-ammonis:
**Fixes #196** 

Confirmed via testing that this does not change add-on behavior. Also ran all tests. Bugzilla seems to [confirm](https://bugzilla.mozilla.org/show_bug.cgi?id=1324255) that the target of this workaround was resolved with FF 60.

### DIFF
--- a/src/popup/modules/UserInterface.js
+++ b/src/popup/modules/UserInterface.js
@@ -24,7 +24,6 @@ import * as QrCreator from "./QrCreator.js";
 import {createMenu} from "/common/modules/ContextMenu.js";
 
 const TOP_SCROLL_TIMEOUT = 10; // ms
-const SELECT_TEXT_TIMEOUT = 100; // ms
 const QR_CODE_REFRESH_TIMEOUT = 200; // ms
 const QR_CODE_CONTAINER_MARGIN = 40; // px
 const QR_CODE_SIZE_SNAP = 5; // px
--- a/src/popup/modules/UserInterface.js
+++ b/src/popup/modules/UserInterface.js
@@ -139,14 +139,8 @@ function isSelected(input) {
  */
 function selectAllText(event) {
     const targetIsSelected = document.activeElement === event.target && isSelected(event.target);
-    // prevent endless loop after two rechecks (i.e. re-check only three times)
-    if (targetIsSelected || event.retry > 3) {
-        return;
-    }
 
     console.info("selectAllText", event);
-
-    event.retry = event.retry + 1 || 0;
 
     // re-selecting when already selected, causes flashing, so we avoid that
     if (!targetIsSelected) {
@@ -180,8 +174,6 @@ function scrollToTop(event) {
         return;
     }
 
-    // Attention: make sure this does not collide with the retry-property set
-    // in selectAllText()!
     event.setScrolled = true;
 }
 
--- a/src/popup/modules/UserInterface.js
+++ b/src/popup/modules/UserInterface.js
@@ -314,23 +314,14 @@ const throttledResizeElements = throttle(resizeElements);
  * Note that this also triggers the actions to show it nicely in the UI.
  *
  * @function
- * @param  {string} text
- * @param  {Object} [options]
- * @param  {boolean} [options.focus=true] focus the text after change
- * @param  {boolean} [options.select=false] select & scroll the text after change
+ * @param  {string} text* *
  * @returns {void}
  */
-export function setQrInputFieldText(text, options = {}) {
+export function setQrInputFieldText(text) {
     qrCodeText.textContent = text;
 
-    // as text has been changed, we need to focus & (potentially) re-select the text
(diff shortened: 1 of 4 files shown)


src/popup/modules/UserInterface.js
```diff
@@ -184,13 +174,7 @@ function scrollToTop(event) {
         return;
     }
 
-    // Attention: make sure this does not collide with the retry-property set
-    // in selectAllText()!
     event.setScrolled = true;
```
rugk: If we don't even re-try scrolling to the top, I guess we can remove all that `setScrolled` logic too, can't we?

rugk:
BTW you can (automatically) [let issues close](https://help.github.com/articles/closing-issues-via-commit-messages/) when a PR is merged with the right keyword (i.e. remove "issue" in your PR description).

src/popup/modules/UserInterface.js
cornu-ammonis: @rugk I agree! But I think there is still a setTimeout for that one [here](https://github.com/cornu-ammonis/offline-qr-code/blob/da9bdeb100dbb3cf028c19ba71dbb3b862a1c2e8/src/popup/modules/UserInterface.js#L153), should we take that out too?

cornu-ammonis:
> Also `SELECT_TEXT_TIMEOUT` is set, but unused now. As per our contributing guide, try to use ESLint.

Ah my bad, will do! Excited to contribute!

src/popup/modules/UserInterface.js
rugk: That's a good question and I am not sure. Is this still needed? Possibly not…?

Or possibly it is, considering "scroll to top" does not work anyway always in the current situation in my testing at least.

cornu-ammonis:
> So I've also changed some things to fix that the selection was not properly triggered when you use the context menu in the add-on.
> 
> Also another issue:
> When you click inside of the input field, it does not always select the whole text automatically:
> gifSelectionIssue
> 
> There is an event listener that should force it to select all the text then: https://github.com/rugk/offline-qr-code/pull/199/files#diff-817dd83d26bd35283a14722780bb8481R516
> 
> Again, it seems this issue is present in the current master, anyway, so it may not be solvable in this PR.

Hmm yeah, does seem like some race stuff. I agree probably out of scope for this PR, but happy to file another issue and see if I can figure it out! Thanks for the comments / additions here!

cornu-ammonis:
> Also `SELECT_TEXT_TIMEOUT` is set, but unused now. As per our contributing guide, try to use ESLint.

removed!

src/popup/modules/UserInterface.js
cornu-ammonis: @rugk yeah, seems like that may call for a larger refactor

rugk:
> happy to file another issue and see if I can figure it out! 

Done: https://github.com/rugk/offline-qr-code/issues/202

src/popup/modules/UserInterface.js
rugk: Okay, fine, accepted. Also opened a different issue for that now: https://github.com/rugk/offline-qr-code/issues/203

rugk:
Ah, BTW, next time, try to avoid doing a pull request from the master branch, because you can run into problems when you have a "non-clean" master that does not follow this repo here (i.e. "upstream"). [See this article for details](https://help.github.com/en/articles/proposing-changes-to-your-work-with-pull-requests).

rugk:
BTW, currently there is the new [Hacktoberfest 2019](https://hacktoberfest.digitalocean.com/) running. Any PR there counts (if it is made in October). :smiley: 
So if you still want to figure out the reason for #202, maybe now is the time, @cornu-ammonis… :wink: 

You may also have a look at [all issues tagged with "hacktoberfest" in this repo](../issues/labels/hacktoberfest) or [somewhere else](https://github.com/issues?utf8=%E2%9C%93&q=is%3Aopen+is%3Aissue+archived%3Afalse+user%3Arugk+user%3ATinyWebEx+label%3A%22hacktoberfest%22+).

PR #201, opened 2019-06-03
close #186 (Automatic) dark mode

### LINKED ISSUE #186: (Automatic) dark mode
## Current experience

image

on GNOME (Fedora/Linux) with Firefox 68 here.

In contrast to Firefox' own/internal popups:
image

### Problems

* A common issue I see on websites (but possibly also an issue of Firefox as it should not handle it like this?), is that input fields inherit the system's white font color by default.
* Generally, the white popup stands out a lot…

## Expected

1. Possibly report the first thing as an issue in Firefox. (don't know how it looks on other OSes)
2. Implement a dark mode that is triggered, when the system uses a dark theme…

To do so, there is this **new fancy** CSS media query: https://developer.mozilla.org/docs/Web/CSS/@media/prefers-color-scheme aka `prefers-color-scheme`
But should the need arise, you can use this JS API: https://developer.mozilla.org/en-US/docs/Web/API/Window/matchMedia

Note: There **must not** be any manual selection of this dark mode. I'd consider this an unnecessary option…

Of course, as always, adhere to the [Photon design](https://design.firefox.com/photon).

3. (optional) As for the QR code itself, that would obviously usually also need to be inverted. So it could be an idea to also change the default colors of the QR code, if it is in dark mode. (of course, do not overwrite user settings though)
  However, this requires some JS changes (which is possible as explained above), but actually even worse: Many QR code scanners read them. (see https://github.com/rugk/offline-qr-code/issues/20)
  So, if we do so, we definitively need to implement https://github.com/rugk/offline-qr-code/issues/20 first. Or find some way that is still scannable, but looks decent in the dark mode.

### Good first issue

This is a good first issue, as it is likely a CSS only change. (And hopefully with not-so many CSS changes.)

Also any more ideas on this are appreciated. :smiley:

davidfloyd91:
Implements a fix to #186: when a user's OS is set to a dark theme, the popup's colors are shifted to match Firefox's dark theme. 

Note that the fix is implemented using `prefers-color-scheme`, which reads from the OS, not the browser. So a user who's chosen a dark macOS theme, but a light Firefox theme, for example, will still see a popup that's formatted to match FF's dark theme.

Also note that if the OS theme is changed, the browser needs to be restarted for offline-qr-code to reflect that change.

### DIFF
--- a/CONTRIBUTORS.md
+++ b/CONTRIBUTORS.md
@@ -3,6 +3,7 @@
 - Alexander Knipping ([@obitech](https://github.com/obitech))
 - Amauri Bizerra ([@AmauriAires](https://github.com/AmauriAires))
 - [@anirudh-modi](https://github.com/anirudh-modi)
+- David Floyd [@davidfloyd91](https://github.com/davidfloyd91)
 - [@ENT8R](https://github.com/ENT8R)
 - Flo Becker ([@beckerei](https://github.com/beckerei))
 - Josh Cooper ([@JoshCooperr](https://github.com/JoshCooperr))
@@ -13,7 +14,6 @@
 - [@redagavin](https://github.com/redagavin)
 - [@Thammarith](https://github.com/Thammarith)
 - [@xlomonosvx](https://github.com/xlomonosvx)
-- David Floyd [@davidfloyd91](https://github.com/davidfloyd91)
 
 ## Translators
 
--- a/CONTRIBUTORS.md
+++ b/CONTRIBUTORS.md
@@ -13,6 +13,7 @@
 - [@redagavin](https://github.com/redagavin)
 - [@Thammarith](https://github.com/Thammarith)
 - [@xlomonosvx](https://github.com/xlomonosvx)
+- David Floyd [@davidfloyd91](https://github.com/davidfloyd91)
 
 ## Translators
 
--- a/CONTRIBUTORS.md
+++ b/CONTRIBUTORS.md
@@ -1,6 +1,7 @@
 # Contributors
 
 - Alexander Knipping ([@obitech](https://github.com/obitech))
+- Andrew Jones ([@cornu-ammonis](https://github.com/cornu-ammonis))
 - Amauri Bizerra ([@AmauriAires](https://github.com/AmauriAires))
 - [@anirudh-modi](https://github.com/anirudh-modi)
 - David Floyd [@davidfloyd91](https://github.com/davidfloyd91)
--- a/src/common/modules/data/DefaultSettings.js
+++ b/src/common/modules/data/DefaultSettings.js
@@ -15,7 +15,7 @@
 // checks for OS dark theme and returns the appropriate background color
  const setQrBackgroundColor = () => {
      if (window.matchMedia('(prefers-color-scheme: dark)').matches) {
-        return "#4a4a4f"; // Photon Grey 60
+        return "#d7d7db"; // Photon Grey 30
      } else {
         return "ffffff";
      }
--- a/src/common/modules/data/DefaultSettings.js
+++ b/src/common/modules/data/DefaultSettings.js
@@ -11,12 +11,22 @@
  * @const
  * @type {Object}
  */
+
+// checks for OS dark theme and returns the appropriate background color
+ const setQrBackgroundColor = () => {
+     if (window.matchMedia('(prefers-color-scheme: dark)').matches) {
+        return "#4a4a4f"; // Photon Grey 60
+     } else {
+        return "ffffff";
+     }
+ };
+
 const defaultSettings = Object.freeze({
     debugMode: false,
     popupIconColored: false,
     qrCodeType: "svg",
     qrColor: "#0c0c0d",
-    qrBackgroundColor: "#ffffff",
+    qrBackgroundColor: setQrBackgroundColor(),
     qrErrorCorrection: "Q",
     autoGetSelectedText: false,
     monospaceFont: false,
--- a/src/popup/modules/UserInterface.js
+++ b/src/popup/modules/UserInterface.js
@@ -536,16 +536,11 @@ export function init() {
             // checks for OS dark theme and sets appropriate background colors
             if (window.matchMedia('(prefers-color-scheme: dark)').matches) {
                 // sets dark theme background and text colors in text field
-                qrCodeText.style.backgroundColor = "#4a4a4f"; // Photon Grey 60
+                qrCodeText.style.backgroundColor = "#4a4a4f"; // Photon Grey 60 -- FF dark theme popup color
                 qrCodeText.style.color = "#ffffff";
-
-                if (qrBackgroundColor === "#4a4a4f") {
-                    // sets qrCode inner background color if user hasn't selected their own color
-                    qrCode.style.backgroundColor = "rgba(249,249,250,0.8)"; // Photon Grey 90 a80
-                } else {
-                    // otherwise matches qrCode inner background color to user's selected background color
-                    qrCode.style.backgroundColor = qrBackgroundColor;
-                }
+            } else {
+                // otherwise matches qrCode inner background color to user's selected background color
+                qrCode.style.backgroundColor = qrBackgroundColor;
             }
         }
     });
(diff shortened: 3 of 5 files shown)


CONTRIBUTORS.md
```diff
@@ -13,6 +13,7 @@
 - [@redagavin](https://github.com/redagavin)
 - [@Thammarith](https://github.com/Thammarith)
 - [@xlomonosvx](https://github.com/xlomonosvx)
+- David Floyd [@davidfloyd91](https://github.com/davidfloyd91)
```
rugk: Please sort it alphabetical, thanks.

src/popup/qrcode.css
```diff
@@ -117,3 +117,17 @@ html, body {
...
   margin: 0px;
 }
+
+/* shift popup colors if user has specified dark mode (reads preference from OS, not browser) */
+@media (prefers-color-scheme: dark) {
+  .qrcode {
+    background-color: var(--grey-30);
+  }
+  #qrcode-container {
+    background-color: var(--grey-60);
```
rugk: Oh another big problem. This change overrides the user's choice of the background color in the settings:
 
image

leads to:

image

---

So I am sorry, but it seems you need to adjust some JS-things, too…
(best solution would IMHO be that it only colors the background of the QR code dark, if you want it.)

So, hmm, I guess we need another change in one module even, where we get the default values from: https://github.com/TinyWebEx/AddonSettings/issues/6

(see also the Readme, it provides an extensive doc on how it's used.)



That color is set in JS.

davidfloyd91:
Sorry about the alphabetizing, I'd read somewhere to add yourself to the bottom. Fixed that.

The colors are from [Photon](https://design.firefox.com/photon/visuals/color.html), yes. The popup background ("Grey 60", #4a4a4f) matches other FF dark theme popups. The QR code background is Photon's "Grey 30" (#d7d7db), which just seemed a bit less blinding than the default white.  

The overwriting is definitely an issue. Just removing the `background-color` from `.qrcode` doesn't appear to work, since there's not enough contrast (see screenshot below). So I'll look at the module you mentioned and fix it in JS.

Screen Shot 2019-06-03 at 2 29 14 PM

rugk:
Okay, so. So marking this as blocked by https://github.com/TinyWebEx/AddonSettings/issues/6

rugk:
>  I'd read somewhere to add yourself to the bottom

You're right, my mistake. Fixed: https://github.com/TinyWebEx/common/commit/90ca64d5cb253b68c82c464b2fa994c79a730331

---

Again regarding this color thing:
* In case I have not expressed it correctly now in a better way: IMHO it would be best if the add-on could automatically "select" dark default colors for the user in case of `prefers-color-scheme = dark`.
* Regarding contrast: Yep, that needs to be assured, by default it should be quite optimal (use the settings page, it checks for that :wink:). Also remember that not all QR code scanners can read inverted QR codes: https://github.com/rugk/offline-qr-code/issues/20
  So that choice of color needs to be a wise one, here… :wink:

davidfloyd91:
Sounds good. I think I've just about wrapped my head around the AddonSettings module, it should be easy enough to add a couple of default settings. I'll keep you posted if I run into issues

rugk:
Yeah, thanks. That sounds great. Feel free to ask any questions then… (best in https://github.com/TinyWebEx/AddonSettings/issues/6, if they are only about that feature there)

davidfloyd91:
Cross-posting this comment, since it's relevant to this issue: https://github.com/TinyWebEx/AddonSettings/issues/6#issuecomment-498772333

rugk:
BTW: Next time, try to avoid doing a pull request from the master branch, because you can run into problems when you have a "non-clean" master that does not follow this repo here (i.e. "upstream"). [See this article for details](https://help.github.com/en/articles/proposing-changes-to-your-work-with-pull-requests).

davidfloyd91:
> BTW: Next time, try to avoid doing a pull request from the master branch, because you can run into problems when you have a "non-clean" master that does not follow this repo here (i.e. "upstream"). See this article for details.

Sorry about that. What's the best way to proceed? I was going to merge the latest commits into master so you can review them, but if that could cause problems I'm happy to close the PR and clean up my master (probably just by re-forking the repo? since we're only talking about 20-something lines of code...)

rugk:
No problem, just keep it as it is now…

src/popup/modules/UserInterface.js
```diff
@@ -532,6 +532,12 @@ export function init() {
     const applyingQrColor = AddonSettings.get("qrBackgroundColor").then((qrBackgroundColor) => {
         if (qrBackgroundColor) {
             document.body.style.backgroundColor = qrBackgroundColor;
+
+            // checks for OS dark theme and sets sets colors in text field
+            if (window.matchMedia('(prefers-color-scheme: dark)').matches) {
+                qrCodeText.style.backgroundColor = "#4a4a4f"; // Photon Grey 60 -- FF dark theme popup color
```
rugk: Why did you move that from CSS to JS now?
IMHO a combination would be just fine…

Everything which you can define statically, put into CSS. (as it was before)
And everything which needs JS-handling, use JS…

src/common/modules/data/DefaultSettings.js
```diff
@@ -11,12 +11,22 @@
...
+     } else {
+        return "ffffff";
+     }
+ };
+
 const defaultSettings = Object.freeze({
     debugMode: false,
     popupIconColored: false,
     qrCodeType: "svg",
     qrColor: "#0c0c0d",
```
rugk: Also generally, I guess you could maybe also adjust the QR code color in case of "dark mode", so that it matches to the background color? Or is this really not needed? (did you check the contrast/tested with QR code scanners etc.)

src/common/modules/data/DefaultSettings.js
```diff
@@ -11,12 +11,22 @@
  * @const
  * @type {Object}
  */
+
+// checks for OS dark theme and returns the appropriate background color
+ const setQrBackgroundColor = () => {
```
rugk: As per [our guide](https://github.com/TinyWebEx/common/blob/master/CONTRIBUTING.md#js):
> Avoid anonymous functions, which have no name (i.e. not really assigned) unless they do really do simple things.

(I've edited the md file now to clarify that a little…)

Generally, don't use arrow functions here. Either inline it into the actual object (I would choose so) or make it a real "usual" function and JSDOC-document it.

rugk: Actually _in this case_ you may even be able to simplify that whole thing to _one_ line: https://developer.mozilla.org/docs/Web/JavaScript/Reference/Operators/Conditional_Operator

src/common/modules/data/DefaultSettings.js
```diff
@@ -11,12 +11,22 @@
...
+     if (window.matchMedia('(prefers-color-scheme: dark)').matches) {
+        return "#d7d7db"; // Photon Grey 30
+     } else {
+        return "ffffff";
```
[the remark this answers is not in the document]
davidfloyd91: Embarrassing. Looks like tests worked because white is the default, I tried it in a ternary with `null` and the result was the same. Fixing for readability

src/popup/modules/UserInterface.js
davidfloyd91: Moving back to CSS

src/common/modules/data/DefaultSettings.js
davidfloyd91: I tried out different combinations, always testing for contrast and QR scannability (if that's a word?). This combination seemed to reduce the white glare, stay in line with Photon and stay scannable better than others. But there might be a better solution

rugk:
Ouch, no, merged the wrong PR…

src/common/modules/data/DefaultSettings.js
davidfloyd91: Sorry, read that in the guide but forgot. Setting to ternary (but pulling `window.matchMedia('(prefers-color-scheme: dark)').matches` out into a `const` so the line doesn't go on forever)

davidfloyd91:
Oh I was wondering. I'll submit one with the changes you mentioned now

davidfloyd91:
Or wait? Let me know what you want to do, here's the commit https://github.com/davidfloyd91/offline-qr-code/commit/1639cef5867d1e88ebc96dbb46cf6afdaee9c515

rugk:
So I reverted master… GitHub does not seem to reopen this PR… so…

Your chance to open a new PR (with a new branch, possibly now :wink:)

Sorry for the inconvenience…

PR #204, opened 2019-06-05
Take two: close #186 (Automatic) dark mode

### LINKED ISSUE #186: (Automatic) dark mode

davidfloyd91:
Fixes https://github.com/rugk/offline-qr-code/issues/186

Incorporates suggested changes from https://github.com/rugk/offline-qr-code/pull/201

### DIFF
--- a/src/common/modules/data/DefaultSettings.js
+++ b/src/common/modules/data/DefaultSettings.js
@@ -19,7 +19,7 @@ const defaultSettings = Object.freeze({
     popupIconColored: false,
     qrCodeType: "svg",
     qrColor: "#0c0c0d",
-    qrBackgroundColor: (darkTheme ? "#d7d7db" : "#ffffff"),
+    qrBackgroundColor: (darkTheme ? "#d7d7db" : "#ffffff"), // dark uses Firefox Photon's grey-30
     qrErrorCorrection: "Q",
     autoGetSelectedText: false,
     monospaceFont: false,
--- a/src/common/modules/data/DefaultSettings.js
+++ b/src/common/modules/data/DefaultSettings.js
@@ -5,7 +5,7 @@
  */
 
 // checks for OS dark theme
-const darkTheme = window.matchMedia('(prefers-color-scheme: dark)').matches;
+const darkTheme = window.matchMedia("(prefers-color-scheme: dark)").matches;
 
 /**
  * An object of all default settings.
--- a/src/common/modules/data/DefaultSettings.js
+++ b/src/common/modules/data/DefaultSettings.js
@@ -15,7 +15,7 @@
 // checks for OS dark theme and returns the appropriate background color
  const setQrBackgroundColor = () => {
      if (window.matchMedia('(prefers-color-scheme: dark)').matches) {
-        return "#4a4a4f"; // Photon Grey 60
+        return "#d7d7db"; // Photon Grey 30
      } else {
         return "ffffff";
      }
--- a/src/common/modules/data/DefaultSettings.js
+++ b/src/common/modules/data/DefaultSettings.js
@@ -4,6 +4,9 @@
  * @module data/DefaultSettings
  */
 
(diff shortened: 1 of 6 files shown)


src/common/modules/data/DefaultSettings.js
```diff
@@ -4,6 +4,9 @@
  * @module data/DefaultSettings
  */
 
+// checks for OS dark theme
+const darkTheme = window.matchMedia('(prefers-color-scheme: dark)').matches;
```
rugk:
```suggestion
const isDarkTheme = window.matchMedia('(prefers-color-scheme: dark)').matches;
```

src/common/modules/data/DefaultSettings.js
```diff
@@ -16,7 +19,7 @@ const defaultSettings = Object.freeze({
     popupIconColored: false,
     qrCodeType: "svg",
     qrColor: "#0c0c0d",
-    qrBackgroundColor: "#ffffff",
+    qrBackgroundColor: (darkTheme ? "#d7d7db" : "#ffffff"),
```
rugk: GitHub makes funny diffs, BTW:
image

Anyway, to get serious, you missed to document that "magic value"=color there now…

rugk:
```suggestion
    qrBackgroundColor: (darkTheme ? "#d7d7db" : "#ffffff"), // dark uses Firefox Photon's grey-30
```

rugk:
```suggestion
    qrBackgroundColor: (isDarkTheme ? "#d7d7db" : "#ffffff"),
```

also obviously

rugk:
Also tested it, here is a sceenshot:
image

So… yeah… looks pretty decent, I guess.

src/common/modules/data/DefaultSettings.js
davidfloyd91: just push this up. couldn't do it here in Github for some reason

src/common/modules/data/DefaultSettings.js
davidfloyd91: pushed this up too

rugk: Thanks yeah, just clicking "resolve conversation" should do it. :smile:

PR #211, opened 2019-07-25
Adjusted the amoSummary.txt file to fit in 250 characters

CGReinhold:
Some other translations were also fixed.


rugk:
Thanks, also for fixing the other translations. :smiley:

Issue #188, opened 2019-04-25
Combine SVG and Canvas generation method


fregante:
1. You can convert the SVG to a canvas by just using `.drawImage(<img src="qr.svg">)`
2. You can always just output SVG

You can offer saving with buttons or a custom context menu

rugk:
So your idea is: Display as an SVG always, for export optionally convert it?

This would indeed simplify the code and CSS a lot and even resizing.
And we could offer SVG and PNG export at the same time.

Only problem I'd have is:
* kjua [by default renders pixel-crisp](https://github.com/lrsjng/kjua/blob/a88369f748926aee8e996e3f720f0f9534c8d86b/src/lib/defaults.js#L5-L6) and you can actually see that when you switch between SVG and Canvas.

rugk:
Actually `qrcodegen.js` also supports Canvas output out-of-the-box, so what is blocking this is really "just" the pixel-perfect rendering.

Thus asked for implementation.

Blocked by https://github.com/nayuki/QR-Code-generator/issues/53

rugk:
I've created a test page for all modes of the QR code libs we use, so that one can have a look whether they look all sharp and suitable: https://rugk.github.io/qr-code-libs-test/

See also discussion at https://github.com/nayuki/QR-Code-generator/issues/53#issuecomment-487510904

rugk:
No need to close, this is still planned.

rugk:
Please stop trolling or I'll have to block you here. If you do not want to receive notifications, click the "unsubscribe" button. Do leave this issue open.

bfred-it:
I don't want this open issue in my list. Open your own issue

rugk:
This is my repo and I will leave this issue open. As said, just click "unsubscribe" and it won't appear in your list.

rugk:
So, finally fixed. So this issue is closed, as you like it… :smiley:

PR #215, opened 2019-08-01
Remove qrcodegen and replace with new kjua-svg version from @werthdavid

rugk:
Fixes https://github.com/rugk/offline-qr-code/issues/188

Uses the new great fork by @werthdavid.


rugk:
Possibly TODO:
* [ ] simplify resizing code, we can now always use SVG and ignore the options for the Canvas

Issue #202, opened 2019-06-05
Text is not always selected clicking into the input field


rugk:
## Steps to reproduce

> When you click inside of the input field, it does not always select the whole text automatically:
>![gifSelectionIssue](https://user-images.githubusercontent.com/11966684/58767460-a2d00180-858b-11e9-8441-3e5bfeebb958.gif)
>
>
> There is an event listener that should force it to select all the text then: https://github.com/rugk/offline-qr-code/pull/199/files#diff-817dd83d26bd35283a14722780bb8481R516

moved from https://github.com/rugk/offline-qr-code/pull/199#pullrequestreview-244642579-body-html

**Actual behavior:**

Often text is not Ctrl+A'ed…

**Expected behavior:**

Always Ctrl+A…

## System

Operation system and version: Linux/Fedora 30/GNOME
Browser and version: Firefox 69
Add-on version: 1.6

winters-brown:
I'll take this up...

PR #220, opened 2019-10-01
Solution To #202

winters-brown:
This simple bug fix should solve the issue of selecting text on click. 

I have no experience in building extensions so I cannot test it as an extension But did verify the solution by loading a live HTML page of the qrcode.html and testing there.

* note I am using windows and the build.sh does not work for me...

### DIFF
--- a/CONTRIBUTORS.md
+++ b/CONTRIBUTORS.md
@@ -15,6 +15,7 @@
 - [@redagavin](https://github.com/redagavin)
 - [@Thammarith](https://github.com/Thammarith)
 - [@xlomonosvx](https://github.com/xlomonosvx)
+- Joshua Winters-Brown [@winters-brown](https://github.com/winters-brown)
 
 ## Translators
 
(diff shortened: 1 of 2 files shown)


rugk:
Hi @winters-brown,

See the doc for how to load/test this add-on: https://github.com/TinyWebEx/common/blob/master/CONTRIBUTING.md#getting-started
As stated there, you do **not** need to run this script.

Anyway, I had a look at your changes and they very likely do not fix the problem, because the event listener and additional select is already implemented: https://github.com/rugk/offline-qr-code/pull/199/files#diff-817dd83d26bd35283a14722780bb8481R516

The issue was just a potential bug in this existing implementation. I also have to say https://github.com/rugk/offline-qr-code/issues/202 is a hard bug (could you actually reproduce it? I.e. did you notice a problem.) That's also why it's tagged with "help wanted".

The issue you are referring actually is not really a great issue for first-timers. For these, we have a different tag named "good first issue". Better [check out these "getting started" issues instead](../contribute).

So I am sorry, but I have to close this PR.

Anyway, one tip for the next time: 
You can (automatically) [let issues close](https://help.github.com/articles/closing-issues-via-commit-messages/) when a PR is merged by adding some "magic" text to your PR _body_.

PR #228, opened 2019-10-24
Fix scrolling to top robustness

### LINKED ISSUE #203: Text input field is not always scrolled to top
## Bug description

Also discovered in https://github.com/rugk/offline-qr-code/pull/199

## Steps to reproduce

1. Go to https://longextendedsubdomainnamewithoutdashesinordertotestwordwrapping.badssl.com/#sdfhhhhhhhhhhhhhhhhhhhhhhhhhhhhhhaklshdiowhaodew902123423423423432erasd
2. Open the QR code popup…

**Actual behavior:**

image

The scroll bar is at the bottom, it's scrolled down.

**Expected behavior:**

Scroll bar should be at the top, i.e. it should display the beginning of the URL.

**More details:**

* **Do also** test different ways to trigger the QR code popup with large data. Ctrl+A and use the context 
* Also test the setting in the options "Automatically use text selected on website.".

The reason that you need to test this is, that all three ways work differently in passing/getting the data to the QR code popup panel. They e.g. trigger at different times and you can expect race conditions or so… 

FYI: opening via hot key is (should be) the same as via click/no difference is expected here.

## System

Operation system and version: Linux/Fedora 30/GNOME
Browser and version: Firefox 69
Add-on version: 1.6

## Further information

Responsible function:
https://github.com/rugk/offline-qr-code/blob/6e817dec0a600171f53665b89b38a62315596d81/src/popup/modules/UserInterface.js#L175

For a solution, I may be fine with refactoring this (it's kinda needed) and possibly split it into it's own file (module). (together with some other text selection things there…)

pradyumnamahajan:
Fixed #203 by making sure scrollToTop was always called and increasing
TOP_SCROLL_TIMEOUT.

setTimeout(scrollToTop, TOP_SCROLL_TIMEOUT, event); was present in an if
statement so it was not always getting called and when it was getting
called, it didn't have enough time to scroll to the top, so to fix this -

1) the setTimeout was moved out of the if statement and

2) TOP_SCROLL_TIMEOUT was increased to 20ms.

### DIFF
--- a/src/popup/modules/UserInterface.js
+++ b/src/popup/modules/UserInterface.js
@@ -149,14 +149,13 @@ function selectAllText(event) {
     // re-selecting when already selected, causes flashing, so we avoid that
     if (!targetIsSelected) {
         event.target.focus();
-        event.target.select();
-
-        // but set scroll position to top one, because you want to see the
-        // top of the URL ;)
-        // (selecting makes the scroll position go to the bottom)
-        
+        event.target.select(); 
     }
     
+    // set scroll position to top one, because you want to see the
+    // start of an URL
+    // (selecting makes the scroll position go to the bottom)
+
     setTimeout(scrollToTop, TOP_SCROLL_TIMEOUT, event);
 }
 
@@ -171,13 +170,6 @@ function selectAllText(event) {
 function scrollToTop(event) {
     console.info("scrollToTop", event); 
     event.target.scrollTo(0,0);
-  
-    // only retry once, if needed
-    if (event.setScrolled) {
-        return;
-    }
-
-    event.setScrolled = true;
 }
 
 /**
--- a/src/popup/modules/UserInterface.js
+++ b/src/popup/modules/UserInterface.js
@@ -23,7 +23,7 @@ import * as QrError from "./QrLib/QrError.js";
 import * as QrCreator from "./QrCreator.js";
 import {createMenu} from "/common/modules/ContextMenu.js";
 
-const TOP_SCROLL_TIMEOUT = 10; // ms
+const TOP_SCROLL_TIMEOUT = 20; // ms
 const QR_CODE_REFRESH_TIMEOUT = 200; // ms
 const QR_CODE_CONTAINER_MARGIN = 40; // px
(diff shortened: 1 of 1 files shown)


Rad0n:
Also all tests ran successfully.
I felt TOP_SCROLL_TIMEOUT = 20ms was enough and always worked for my machine but if the problem is still present, I can increase the timeout even more.

I also felt that 
`
    // only retry once, if needed
    if (event.setScrolled) {
        return;
    }

    event.setScrolled = true;
`
can be removed because its not being used anywhere else in the file.

Rad0n:
Also all tests ran successfully.
I felt TOP_SCROLL_TIMEOUT = 20ms was enough and always worked for my machine but if the problem is still present, I can increase the timeout even more.

I also felt that 
`
    // only retry once, if needed
    if (event.setScrolled) {
        return;
    }

    event.setScrolled = true;
`
can be removed because its not being used anywhere else in the file.

src/popup/modules/UserInterface.js
```diff
@@ -154,8 +154,10 @@ function selectAllText(event) {
         // but set scroll position to top one, because you want to see the
```
rugk: Could you also adjust the comment then or at least move it to the correct position? Thanks.

src/popup/modules/UserInterface.js
```diff
@@ -167,12 +169,9 @@ function selectAllText(event) {
...
-
-    if (event.target.scrollTop !== 0) {
-        event.target.scrollTop = 0;
-    }
-
+    console.info("scrollToTop", event); 
+    event.target.scrollTo(0,0);
+  
     // only retry once, if needed
     if (event.setScrolled) {
```
rugk: Mhh, it seems I've introduced this, because before, here was a loop or so, that passed on the `event` parameter. I just guess, now it does not do that anymore. So as far as I see, it may indeed be useless, because it is first sentence, then the whole thing is discarded and the new event is a different one, when we get again here…

src/popup/modules/UserInterface.js
Rad0n: yeah, will do.

PR #250, opened 2020-06-02
French translations fixes

dvergeylen:
Using the polite form ('vous') rather than the familiar form ('tu'), which sounds childish to the ear.
Also updated the 1.7 changelog.
Thank you for this extension! 🏄

### DIFF
--- a/assets/texts/fr/Changelog/1.7.html
+++ b/assets/texts/fr/Changelog/1.7.html
@@ -1,12 +1,12 @@
 <ul>
-<li><strong>Nouveau :</strong> Grâce à une nouvelle librarie de génération de QR Code, vous pouvez maintenant toujours sauvegarder le QR comme PNG ou SVG. Le paramètre pour choisir entre ces deux modes a été retiré des options. (<a href="https://github.com/rugk/offline-qr-code/issues/188">#188</a>)</li>
-<li><strong>Nouveau :</strong> L'extension supporte maintenant le mode "dark" du système. (<a href="https://github.com/rugk/offline-qr-code/issues/186">#186</a>, <a href="https://github.com/rugk/offline-qr-code/issues/190">#190</a>, merci à <a href="https://github.com/davidfloyd91">@davidfloyd91</a>)</li>
-<li><strong>Nouveau :</strong> La traduction vers le chinois (traditionnel) a été ajoutée, grâce à <a href="https://github.com/niotw">@Nio</a>.</li>
-<li><strong>Nouveau :</strong> La traduction vers l'hébreux a été ajoutée, grâce à <a href="https://github.com/shanirub">@shanirub</a>.</li>
-<li><strong>Nouveau:</strong> La traduction vers le japonais a été ajoutée, grâce à <a href="https://github.com/fluoros-jp">@fluoros-jp</a>.</li>
-<li><strong>Nouveau:</strong> La traduction vers le portugais a été ajoutée, grâce à <a href="https://github.com/CGReinhold">@CGReinhold</a>.</li>
-<li><strong>Nouveau:</strong> La traduction vers le russe a été ajoutée, grâce à <a href="https://github.com/m1ndm1n3r">@m1ndm1n3r</a>.</li>
-<li><strong>Nouveau:</strong> La traduction vers le slovaque a été ajoutée, grâce à <a href="https://github.com/daaAd1">@daaAd1</a>.</li>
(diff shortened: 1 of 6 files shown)


assets/texts/fr/Changelog/1.7.html
```diff
@@ -0,0 +1,12 @@
+<ul>
+<li><strong>Nouveauté :</strong> Grâce à une nouvelle librarie de génération de QR Code, vous pouvez maintenant toujours sauvegarder le QR comme PNG ou SVG. Le paramètre pour choisir entre ces deux modes a été retiré des options. (<a href="https://github.com/rugk/offline-qr-code/issues/188">#188</a>)</li>
```
nicolas-trouin, line 2: I suggest using the correct translation for "library", "bibliothèque", instead of false friend "librairie" which means "book shop" even if those are often mixed up.

assets/texts/fr/Changelog/1.7.html
```diff
@@ -0,0 +1,12 @@
...
+<li><strong>Nouveauté :</strong> L'extension supporte maintenant le mode "dark" du système. (<a href="https://github.com/rugk/offline-qr-code/issues/186">#186</a>, <a href="https://github.com/rugk/offline-qr-code/issues/190">#190</a>, merci à <a href="https://github.com/davidfloyd91">@davidfloyd91</a>)</li>
```
nicolas-trouin, line 3: I suggest replacing 
> mode "dark"

with somehting like
> mode sombre (dark)

assets/texts/fr/Changelog/1.7.html
```diff
@@ -0,0 +1,12 @@
...
+<li><strong>Nouveauté :</strong> La traduction vers le chinois (traditionnel) a été ajoutée, grâce à <a href="https://github.com/niotw">@Nio</a>.</li>
```
nicolas-trouin, line 4: I would rather use "en" instead of "vers le", for example : 
> La traduction en chinois (traditionnel) ...

or
> La traduction en hébreux

src/_locales/fr/messages.json
```diff
@@ -120,24 +120,24 @@
     "hash": "a2d1d2fdc32ecb73ffdf0ab537e46f32"
   },
   "messageAutoSelectColorButton": {
-    "message": "Utilise des couleurs en contraste.",
+    "message": "Utilisez des couleurs contrastées.",
```
nicolas-trouin, line 123: Not sure about what is this should mean, so both "Utilise" and "Utilisez" (as well as "Utiliser") *could* be correct. Maybe @rugk can give a bit more of context ?

nicolas-trouin: I agree with "contrastées" 👍

nicolas-trouin:
Also nice to see some other people help with French translation :)

src/_locales/fr/messages.json
dvergeylen: I also struggled to understand the sentence before I read it in the addon itself.
Go to addon preferences and select two same colors (front and background), you will see the error message :wink:

dvergeylen: > I agree with "contrastées" +1

:+1:

assets/texts/fr/Changelog/1.7.html
dvergeylen: Hum, 'librairie' is used far more often but I won't be stubborn here if you insist to use 'bibliothèque' instead 🙄

src/_locales/fr/messages.json
rugk: > The text of a button that sets a new color with a sufficient contrast.

It#s a button, you can click on. So it should be imperative.

https://github.com/rugk/offline-qr-code/blob/a6eba310a10b2fb003a34d3938587b9ad5d273b1/src/_locales/en/messages.json#L103-L106

rugk:
I totally agree. Good folks here! :smiley: 

So, if everything is all right to the reviewer(s) aka @nicolas-trouin, please use the GitHub "approve" feature to approve the changes.

assets/texts/fr/Changelog/1.7.html
nicolas-trouin: I know "librairie" is widely used but it remains "uncorrect". French speaking people with knowledge in computer science will understand both terms.
Well at least you should correct "librairie", you forgot an "i".

dvergeylen:
- Translated 'librarie' to 'bibliothèque'
- 'Utilisez' → 'Utiliser' as rugk specified it's a button

rugk:
Okay, merged and deployed on AMO.

PR #252, opened 2020-06-07
Filename generation for URL

### LINKED ISSUE #247: Set download name to domain name for URLs
First thank you very much for developing this wonderful and very helpful Extension 👍 🥇 

## Background
Storing  multiple QR codes from URL's is tedious, since you want to rename them sensible, to know what is in there.

## Proposed solution
If we have an URL pattern :// I would like to have the download name set to the domain name (or host name if it is easier) with '.' substituted by '_' .

Is this possible?
Makes it sense for me to develop a change and issue an PR?

Greetings from Cologne
Armin

ArminJo:
As discussed before.
Pretty printing may be required.
Closes #247

### DIFF
--- a/src/popup/modules/UserInterface.js
+++ b/src/popup/modules/UserInterface.js
@@ -52,6 +52,7 @@ let placeholderShown = true;
 let qrLastSize = 200;
 let qrCodeSizeOption = {};
 let savingQrCodeSize = null; // promise
+let qrCodeInputText = "";
 
 /**
  * Hide QR code and show placeholder instead.
@@ -103,7 +104,8 @@ function hidePlaceholder() {
 const refreshQrCode = throttle(() => {
     const text = qrCodeText.value;
     console.info("new value from textarea: ", text);
-
+    qrCodeInputText = text; // Save it for filename generation
+    
     // show placeholder when no text is entered
     if (text === "") {
         showPlaceholder();
@@ -436,6 +438,37 @@ function triggerFileSave(file, filename, requestDownloadPermissions) {
 }
 
 /**
+ * If qrcode text was an URL i.e contains :// this function generates a filename using the domain and replacing all special characters with _
+ *.If not return "qrcode"
+ *
+ * @function
+ * @private
+ * @param { } 
+ * @returns {filename}
+ */
+function generateFilename(){
+	var firstIndex = qrCodeInputText.indexOf("://");
+	if(firstIndex > 0) {
+		// URL found
+		var filename = qrCodeInputText.substring(firstIndex+3);
+		var wwwIndex = filename.indexOf("www");
+		if(wwwIndex >= 0){
+			// remove www(*.).
+			filename = filename.substring(filename.indexOf(".")+1);
+		}
+		// keep text until first "/"
+		filename = filename.substring(0,filename.indexOf("/"));
+		// replace . by _
+		//filename = filename.replace(/\./g,"_");
+		// Replace all strange characters
+		filename = filename.replace(/[^a-z0-9_-]/g,"_");
+		// prepend "qrcode"
+		return "qrcode-" + filename;			
+	}
+	return "qrcode";
+}
+
+/**
  * Triggers when a context menu item has been clicked.
  *
  * It downloads the QR code image.
@@ -447,20 +480,23 @@ function triggerFileSave(file, filename, requestDownloadPermissions) {
  */
 function menuClicked(event) {
     const requestDownloadPermissions = browser.permissions.request(DOWNLOAD_PERMISSION);
-
+    var filename = generateFilename();
+    
     switch (event.menuItemId) {
     case CONTEXT_MENU_SAVE_IMAGE_SVG: {
+        filename= filename + ".svg";
         const svgElem = QrCreator.getQrCodeSvgFromLib();
         const svgString = (new XMLSerializer()).serializeToString(svgElem);
-        const file = new File([svgString], "qrcode.svg", { type: "image/svg+xml;charset=utf-8" });
-        triggerFileSave(file, "qrcode.svg", requestDownloadPermissions);
+        const file = new File([svgString], filename, { type: "image/svg+xml;charset=utf-8" });
+        triggerFileSave(file, filename, requestDownloadPermissions);
         break;
     }
     case CONTEXT_MENU_SAVE_IMAGE_CANVAS: {
+        filename= filename + ".png";
         const canvasElem = QrCreator.getQrCodeCanvasFromLib();
         canvasElem.toBlob((blob) => {
-            const file = new File([blob], "qrcode.png", { type: "image/png" });
-            triggerFileSave(file, "qrcode.png", requestDownloadPermissions);
+            const file = new File([blob], filename, { type: "image/png" });
+            triggerFileSave(file, filename, requestDownloadPermissions);
         }, "image/png");
         break;
     }
--- a/src/popup/modules/UserInterface.js
+++ b/src/popup/modules/UserInterface.js
@@ -447,20 +447,18 @@ function triggerFileSave(file, filename, requestDownloadPermissions) {
  * @returns {filename}
  */
 function generateFilename(){
-	var firstIndex = qrCodeInputText.indexOf("://");
+	let firstIndex = qrCodeInputText.indexOf("://");
 	if(firstIndex > 0) {
 		// URL found
-		var filename = qrCodeInputText.substring(firstIndex + 3);
-		var wwwIndex = filename.indexOf("www");
+		let filename = qrCodeInputText.substring(firstIndex + 3);
+		let wwwIndex = filename.indexOf("www");
 		if (wwwIndex >= 0) {
 			// remove www(*.).
-			filename = filename.substring(filename.indexOf(".")+1);
+			filename = filename.substring(filename.indexOf(".") + 1);
 		}
 		// keep text until first "/"
 		filename = filename.substring(0,filename.indexOf("/"));
-		// replace . by _
-		//filename = filename.replace(/\./g,"_");
-		// Replace all strange characters
+		// Replace "." and all strange characters by _
 		filename = filename.replace(/[^a-z0-9_-]/g,"_");
 		// prepend "qrcode"
 		return "qrcode-" + filename;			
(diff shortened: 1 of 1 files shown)


rugk:
I'll have a look at this PR soon.

You can (automatically) [let issues close](https://help.github.com/articles/closing-issues-via-commit-messages/) when a PR is merged by adding some "magic" text to your PR _body_. It would be kind, if you could edit your PR to change this.

Also please do also note one _important_ precautionary warning of me: In case you did not notice, you have committed (i.e. created commits) with potentially sensitive data like your (full/private) name or mail.
Anyone, who clones your fork/this repo can view these private details, exactly as you've committed them. Also, if this PR gets merged, this will stay in the (commit) history of this project and will thus hard to be removed later.

So if this was unintentional and you want that data to be changed/removed first, better do it now. :smiley: [Here is a good resource](https://stackoverflow.com/a/750182/5008962) ([actually from GitHub's help](https://help.github.com/en/articles/changing-author-info)) on how to change your name/mail afterwards.
Also note that when setting your git [username](https://help.github.com/en/articles/setting-your-username-in-git) and [especially your email](https://help.github.com/en/articles/setting-your-commit-email-address), you can actually write/choose anything there. As for GitHub, it is recommend to use your GitHub username and – if you want to keep your mail private – the GitHub provided `noreply` address. You can [even block pushes of commits with private mails](https://help.github.com/en/articles/blocking-command-line-pushes-that-expose-your-personal-email-address).

So far just the possibilities. I just wanted to serve this as a friendly reminder, so you don't notice this only when it's too late. What you'll do depends solely on you. :smile: 

---

BTW: Next time, try to avoid doing a pull request from the master branch, because you can run into problems when you have a "non-clean" master that does not follow this repo here (i.e. "upstream"). [See this article for details](https://help.github.com/en/articles/proposing-changes-to-your-work-with-pull-requests). Anyway, this only applies the next time.

src/popup/modules/UserInterface.js
```diff
@@ -435,6 +437,37 @@ function triggerFileSave(file, filename, requestDownloadPermissions) {
...
+ * @param { } 
+ * @returns {filename}
+ */
+function generateFilename(){
+	var firstIndex = qrCodeInputText.indexOf("://");
+	if(firstIndex > 0) {
+		// URL found
+		var filename = qrCodeInputText.substring(firstIndex+3);
+		var wwwIndex = filename.indexOf("www");
+		if(wwwIndex >= 0){
```
rugk, line 455:
```suggestion
		if (wwwIndex >= 0) {
```

spaces please

src/popup/modules/UserInterface.js
```diff
@@ -435,6 +437,37 @@ function triggerFileSave(file, filename, requestDownloadPermissions) {
...
+ * @function
+ * @private
+ * @param { } 
+ * @returns {filename}
+ */
+function generateFilename(){
+	var firstIndex = qrCodeInputText.indexOf("://");
+	if(firstIndex > 0) {
+		// URL found
+		var filename = qrCodeInputText.substring(firstIndex+3);
```
rugk, line 453:
```suggestion
		var filename = qrCodeInputText.substring(firstIndex + 3);
```

As per our contributing guide, please use EsLint that checks/enforces such code style things…

src/popup/modules/UserInterface.js
```diff
@@ -447,20 +480,23 @@ function triggerFileSave(file, filename, requestDownloadPermissions) {
  */
 function menuClicked(event) {
     const requestDownloadPermissions = browser.permissions.request(DOWNLOAD_PERMISSION);
-
+    var filename = generateFilename();
```
rugk, line 483: Don't use `var`, that's outdated. Again, eslint should tell you to use const or let.

src/popup/modules/UserInterface.js
```diff
@@ -435,6 +437,37 @@ function triggerFileSave(file, filename, requestDownloadPermissions) {
...
+ *.If not return "qrcode"
+ *
+ * @function
+ * @private
+ * @param { } 
+ * @returns {filename}
+ */
+function generateFilename(){
+	var firstIndex = qrCodeInputText.indexOf("://");
+	if(firstIndex > 0) {
```
rugk, line 451: Also per our contributing guide: use "early return" if styles, i.e. check for the inverse result and return the fallback.

src/popup/modules/UserInterface.js
```diff
@@ -435,6 +437,37 @@ function triggerFileSave(file, filename, requestDownloadPermissions) {
...
+ * If qrcode text was an URL i.e contains :// this function generates a filename using the domain and replacing all special characters with _
+ *.If not return "qrcode"
+ *
+ * @function
+ * @private
+ * @param { } 
+ * @returns {filename}
+ */
+function generateFilename(){
+	var firstIndex = qrCodeInputText.indexOf("://");
```
rugk, line 450: Yeah `indexOf` is possible. Just to let you know: [JS has newer methods like `.includes`](https://developer.mozilla.org/docs/Web/JavaScript/Reference/Global_Objects/String/includes) that are more intuitive to read and maybe using that here would be better?

rugk: Or wait… [`startsWith`](https://developer.mozilla.org/docs/Web/JavaScript/Reference/Global_Objects/String/startsWith) would be even better? Cause we don't want to return the domain only, if the input e.g. is `Check out the URL: https://test.com` the qr code would be named `test` or so, which would be bad IMHO.

So rather be more strict here…

ArminJo:
Thanks for the remainder.

src/popup/modules/UserInterface.js
```diff
@@ -435,6 +437,37 @@ function triggerFileSave(file, filename, requestDownloadPermissions) {
...
+		// keep text until first "/"
+		filename = filename.substring(0,filename.indexOf("/"));
+		// replace . by _
+		//filename = filename.replace(/\./g,"_");
+		// Replace all strange characters
+		filename = filename.replace(/[^a-z0-9_-]/g,"_");
+		// prepend "qrcode"
+		return "qrcode-" + filename;			
+	}
+	return "qrcode";
```
rugk, line 468: Generally, as for the fallback, don't you think it could still make sense to return the first, say 10 chars?

src/popup/modules/UserInterface.js
```diff
@@ -435,6 +437,37 @@ function triggerFileSave(file, filename, requestDownloadPermissions) {
...
+		var filename = qrCodeInputText.substring(firstIndex+3);
+		var wwwIndex = filename.indexOf("www");
+		if(wwwIndex >= 0){
+			// remove www(*.).
+			filename = filename.substring(filename.indexOf(".")+1);
+		}
+		// keep text until first "/"
+		filename = filename.substring(0,filename.indexOf("/"));
+		// replace . by _
+		//filename = filename.replace(/\./g,"_");
```
rugk, line 462: so why is this uncommented? Shall we include it now or not?

IMHO at least Windows is quite sensitive with too many dots in your filename. So IMHO replace it. (But line 463 should also already do that, should not it?)

src/popup/modules/UserInterface.js
ArminJo: What, if they consist of special characters.
Than you have a lot of underscores in the fiename.
The URL syntax is quite restricted, but any free text not.

src/popup/modules/UserInterface.js
ArminJo: This works also for http:// and ftp:// so why change?

ArminJo: I want to concentrate on the basics and 95% of the assumed usage and make it KISS.
The user still has the possibility to change the filename in the system download dialog.

src/popup/modules/UserInterface.js
ArminJo: Sorry I learned that early return is bad style, but this was a few years ago.
But of course I can change this if you prefer.

ArminJo: I intended it as the "error" return.

src/popup/modules/UserInterface.js
ArminJo: It is replaced by the replace()

ArminJo:
Hi, I am doing my first review, so I resolved each conversation, but I do not know the exact semantics of "resolving", so you must apologize, if I made an error.

src/popup/modules/UserInterface.js
rugk: Well… it's stated in the contribution guide, so please follow what is written there. That does, of course, not mean these rules are written into stone and may never be changed, but if we'd want to change this, that would be a thing we need to discuss first.
As for this case, I think – as usual – it is a matter of taste and possibly depends on other factors like the used programming language etc. As for this project, the “official taste” is to use early return.  The [Stackexchange discussion](https://softwareengineering.stackexchange.com/q/18454/209137) for this feature also lists some points on why this is useful, if you want to see some reasoning for that rule.

src/popup/modules/UserInterface.js
rugk: Okay, valid argument.

src/popup/modules/UserInterface.js
ArminJo: So you want to have 2 error exits with the same return values?

ArminJo:
Hello,
1. Thank you for the flowers and the medal 😀 
2. do you know this Dilbert?
Dilbert_Plan
3. My last [Javascript program](http://www.joachimsmeyer.de/src/all_example.htm) was 11 years ago.
4. [Eslint online](https://eslint.org/demo) only shows: <button class="alert btn-block report-message-alert alert-danger" title="'generateFilename' is defined but never used.">2:10 - 'generateFilename' is defined but never used. (<a href="https://eslint.org/docs/rules/no-unused-vars" target="_blank" rel="noopener noreferrer">no-unused-vars</a>)</button>
Or if I use let:
<button class="alert btn-block report-message-alert alert-danger" title="Parsing error: Unexpected token qrCodeInputText">1:5 - Parsing error: Unexpected token qrCodeInputText</button>

5. If you have a question regarding [Arduino ](https://cse.google.com/cse?cx=008830049033802121810%3Azgsyidwem5w&q=joachimsmeyer)I can help you, so please help me and send me the corrected generateFilename() function, I will take it as you send.

rugk:
Here I've [changed it](https://github.com/rugk/offline-qr-code/compare/37c70ed1b999d29599f645fc3c317acd82d5dbe5...eb66a275ad19e0154d79bb66d6bf795488b9fae2) and merged some thing.

Please have a few looks at the commits (at the end of that list there).

Basically I've tried to handle more side-effects for URL variants, and in the end discovered that browsers actually have an integrated URL parsing mechanism for JS, so I've switched to this.

ArminJo:
Thank you very much! 👍 👍 
I have learned a lot from your commits!

PR #258, opened 2020-09-30
Update README.md

tanaychaulinsec:


rugk:
Hmm, this is no legal disclaimer. Also a QR code can be used for multiple things.
I don't see a reason to add that string there. The reason for that trademark string was just a legal one. 🙂

PR #283, opened 2022-09-19
Added unit tests

atharvarajurkar:


src/tests/lodashTest/isObjectLike.test.js
```diff
@@ -0,0 +1,31 @@
+import "https://unpkg.com/mocha@5.2.0/mocha.js"; /* globals mocha */
+import "https://unpkg.com/chai@4.1.2/chai.js"; /* globals chai */
+
+import isObjectLike from "/common/modules/lodash/isObjectLike.js";
+
+describe("common lodash: isObjectLike", function () {
```
rugk, line 6: That's a very good intent, but I doubt we need to test lodash modules again. They are teted by lodash already and may/are in the future even considered to be dropped: https://github.com/TinyWebEx/AddonTemplate/issues/7

But sure this was a good test to do,

Same goes for the other files. `BrowserCommunicationTypes` is a thing I would keep, the other ones are kinda redundant.

rugk (changes requested):
Hi @atharvarajurkar,

Did you actually run these tests according to the contribution guidelines? (As these are unfortunately quite hard to run…) If so a screenshot of that would be nice.

But before you do the work, see my other comment, as I assume you can delete the lodash files again. We do not need to test stuff that is already tested.

Anyway, I hope this was a good practice and you've learned unit testing things.

Issue #282, opened 2022-07-29
Use system clipboard content than tab url to gen code.


qrqhuang:
I was commonly used clipboard content to generate code.
Try to give an option to choose.

rugk:
Good idea. Note that this [would require a new permission](https://developer.mozilla.org/en-US/docs/Mozilla/Add-ons/WebExtensions/Interact_with_the_clipboard) for clipboard access likely, but that can be queried

## HowTo

I imagine a setting with a checkbox:
* [x] Use clipboard content when opening popup.

If you enable that, instead of [using the tab URL](https://github.com/rugk/offline-qr-code/blob/b15ecc86a4d386dbc08de868ea773c649b921ca1/src/popup/modules/InitQrCode.js#L78-L79) it should override it (if required also [possible with a message from the background script or so](https://github.com/rugk/offline-qr-code/blob/b15ecc86a4d386dbc08de868ea773c649b921ca1/src/popup/modules/ReceiveBackgroundMessages.js#L25) with the clipboard content.
If a permission is required, one can couple it in this extension [with the PermissionRequest](https://github.com/TinyWebEx/PermissionRequest) library (also used in other extensions, if one needs to get an example).

Important things:
* If the clipboard is empty, it should IMHO still fallback/use the website URL then as usual!
* If some text is selected and the option in the settings is enabled (as usual), this should take preference over the clipboard content IMHO.
  **Reasoning:** The thing is the clipboard is likely to contain outdated stuff, even passwords or so, which you have copied a quite long time ago, while a text selection on the website is _much_ less permanent (i.e. goes away fast when you have to interact in some other way with the site) _and_ selecting something may be a way more deliberate act than just copying sth. in another application even possibly or so.
  
That all said, I guess this is a good „good first issue“ thing if someone wants to really implement it. Note though, it may be not the easiest one, as one has to touch a few topics (settings, actual logic etc.) that are simplified due to the libs we use, but may still require some clever ideas or so.

frooia:
New open source dev here, can I try this one?

Ravina-Deogadkar:
Can I work on this?

rugk:
Hi oh sorry for the delay, missed my notifications, yes you can both try. However, @Ravina-Deogadkar I just saw @frooia already did a PR in https://github.com/rugk/offline-qr-code/pull/286, so that is the base for implementation.

[...]

PR #286, opened 2022-10-02
Clipboard content can be used to generate QR code

frooia:
This pull request attempts to provide a solution to this [issue](https://github.com/rugk/offline-qr-code/issues/282). It introduces a new permission called `clipboardRead`, which is necessary for reading content from the user's clipboard. It has been documented in the permissions README. An option has been added under Preferences so that the user can select if they wish to have clipboard content automatically used to generate the QR code, rather than tab URL.


frooia:
Following up on this issue, is someone able to test and review it? Thanks.

src/manifest.json
```diff
@@ -52,7 +52,8 @@
   "permissions": [
     "activeTab",
     "storage",
-    "menus"
+    "menus",
+    "clipboardRead"
```
rugk, line 56: Can you make it an optional permission (see `optional_permissions`), so we don't request that permission from anyone?

See what I wrote before:
> If a permission is required, one can couple it in this extension [with the PermissionRequest](https://github.com/TinyWebEx/PermissionRequest) library (also used in other extensions, if one needs to get an example).

https://github.com/rugk/offline-qr-code/issues/282#issuecomment-1199975970

As a comparison, see [how the awesome emoji picker implements this](https://github.com/rugk/awesome-emoji-picker/blob/26295d9b391ee6a48b2e636f1339e109989f5f44/src/options/modules/CustomOptionTriggers.js#L317-L334) for clipboard pasting in that case:

image

src/options/options.html
```diff
@@ -172,6 +172,13 @@ <h1 data-i18n="__MSG_titleAddonBehavior__">Addon behavior</h1>
 						</div>
 					</li>
 
+					<li>
+						<div class="line">
+							<input class="setting save-on-change" type="checkbox" id="autoGetClipboardContent" name="autoGetClipboardContent">
+							<label data-i18n="__MSG_optionAutoGetClipboardContent__" for="autoGetClipboardContent">Use clipboard content when opening popup</label>
```
rugk, line 178: Yeah good, now only the `optionAutoGetClipboardContent` also needs localization in the `message.json` file. English is fine for now (just duplicate the string there, too).

See https://developer.mozilla.org/en-US/docs/Mozilla/Add-ons/WebExtensions/Internationalization#providing_localized_strings_in__locales for more information.

assets/texts/en/permissions.md
```diff
@@ -17,8 +17,9 @@ These permissions are requested when doing some specific actions, if they are ne
...
-|:------------|:-----------------------------|:------------------------------------------------------------------|
-| `activeTab` | Access current tab website   | Needed for getting the URL of the current website for the QR code |
-| `storage`   | Access local storage         | Needed for saving options                                         |
-| `menus`     | Modify browser context menus | Needed for adding context menus "QR code from selection" (etc.)   |
+| Internal Id    | Permission                   | Explanation                                                       |
+|:---------------|:-----------------------------|:------------------------------------------------------------------|
+| `activeTab`    | Access current tab website   | Needed for getting the URL of the current website for the QR code |
+| `storage`      | Access local storage         | Needed for saving options                                         |
+| `menus`        | Modify browser context menus | Needed for adding context menus "QR code from selection" (etc.)   |
+| `clipboardRead`| Access clipboard content     | Needed for using clipboard content to create QR code              |
```
rugk, line 25: Thanks for doing the doc part, too! However, currently it is in the wrong section, as currently it would be an "Installation permissions" and actually _not_ be hidden.

_However_, before you change that, see my suggestion below. Actually I'd like if it were an optional permission, so this would be placed under "Feature-specific (optional) permissions".

src/popup/modules/InitQrCode.js
```diff
@@ -56,6 +56,14 @@ const gettingSelection = AddonSettings.get("autoGetSelectedText").then((autoGetS
     });
 });
 
+// check for clipboard text
+const gettingClipboard = AddonSettings.get("autoGetClipboardContent").then((autoGetClipboardContent) => {
+    if (autoGetClipboardContent !== true) {
+        return Promise.reject(new Error("using clipboard content is disabled"));
+    }
+    return navigator.clipboard.readText();
```
rugk, line 64: [I previously outlined one import test case](https://github.com/rugk/offline-qr-code/issues/282#issuecomment-1199975970):
> If the clipboard is empty, it should IMHO still fallback/use the website URL then as usual!

Did you test that?
If it does not work, could you please adjust it to do that? (I imagine you could just check the return value of that API here and check if it's empty and reject the Promise if needed.)

rugk (changes requested):
Hi @frooia,

In general, I very much like implementation, it seems simple and should work.
However making it optional and [some minor adjustments to the acceptance criteria I outlined before](https://github.com/rugk/offline-qr-code/issues/282#issuecomment-1199975970) would be needed.

Of course, you would finish the whole thing if you could also include a Changelog entry (English is of course enough), to be found in the `assets` directory.

BTW: Next time, try to avoid doing a pull request from the master branch, because you can run into problems when you have a "non-clean" master that does not follow this repo here (i.e. "upstream"). [See this article for details](https://help.github.com/en/articles/proposing-changes-to-your-work-with-pull-requests). Anyway, this is only a tip for the next time. :smiley:

assets/texts/en/permissions.md
rugk: See https://extensionworkshop.com/documentation/develop/request-the-right-permissions/#request-permissions-at-runtime

rugk:
@frooia  Are you still interested in it? If not, that's no problem, just add a notice, so I know what the current status of this issue is.

PR #290, opened 2022-10-22
Grammatical error in Readme file

pulkitkhanna1:
This is a fix to the grammatical errors in the readme file which may have occurred while translating from native language, this is under the HACKTOBER FEST 2022


rugk (changes requested):
These are no grammatical errors and also some superflous words. A QR code cannot be dynamic and it's obvious you can modify the text if one says you can enter _any_ text… hmm?

Issue #189, opened 2019-04-25
Allow resizing via (Ctrl+)mousewheel + pinch-to-zoom


rugk:
> I'd suggest resizing the QR via mousewheel event instead of the textarea. 

by @bfred-it  in https://github.com/rugk/offline-qr-code/issues/57#issuecomment-486772326

That's especially useful for touch devices.

---

Maybe we can intercept the page zoom on Ctrl+scroll wheel, that currently "works" i.e. leads to:
![image](https://user-images.githubusercontent.com/11966684/56758216-6512d880-6796-11e9-869e-b65f6569400c.png)

eridilla:
Hi, could I work on this?

rugk:
@eridilla Sure, you can. The UX needs some testing here, i.e. IÄm not sure what is most useful.
For a first try I would say 
* "<kbd>Ctrl</kbd>+mouse wheel" should resize
* "<kbd>Ctrl</kbd>+plus/minus key" should resize

So I just tested the current state and while + and - key do not work (the last two points) - it just resizes the tab I have open, it seems that mouse wheel already works. It sometimes resizes the tab/website, too, but it certainly also zooms

Maybe though:
* that zoom level should possibly be remembered and used/applied when opening the next time, so that you have a persistent zoom (via this mode) and a temporary (via the corner that works as usual, currently) 
  This is similar to how Firefox treats the zoom by default: each website remembers it's zoom level.
  
Technically, you may experiment between:
* the current way: changing the size of the generated QR code
* and some real "zoom" level setting/API: https://developer.mozilla.org/en-US/docs/Mozilla/Add-ons/WebExtensions/API/Tabs/setZoom, this seems [rather impossible in plain JS though](https://stackoverflow.com/q/1055336/5008962)

Firefox 105.0.2 (64-bit) on Linux/Fedora

eridilla:
@rugk Hey, I have implemented the zoom function using the keyboard, but I thought that it's generally not a good idea to overwrite basic key binds like "Ctrl + plus/minus" and also when I tried, I haven't found a way that made it only zoom the extension and not the whole tab, so instead I made it so it listens to "Alt + plus/minus" and added a little tooltip for it. The zoom levels have a boundary too so the generated code doesn't move in the container and stays centered. The zoom also persists after reopening the extension like you mentioned in your reply.

Regarding the pinch zoom, I don't have a means to test that function if I would work on it, so I haven't looked at that.

[...]

PR #293, opened 2022-12-18
Keyboard zoom

eridilla:
Issue #189

### DIFF
--- a/src/popup/modules/UserInterface.js
+++ b/src/popup/modules/UserInterface.js
@@ -21,7 +21,7 @@ import * as CommonMessages from "/common/modules/MessageHandler/CommonMessages.j
 
 import * as QrError from "./QrLib/QrError.js";
 import * as QrCreator from "./QrCreator.js";
-import {createMenu} from "/common/modules/ContextMenu.js";
+import { createMenu } from "/common/modules/ContextMenu.js";
 
 const TOP_SCROLL_TIMEOUT = 20; // ms
 const QR_CODE_REFRESH_TIMEOUT = 200; // ms
@@ -35,13 +35,15 @@ const CONTEXT_MENU_SAVE_IMAGE_CANVAS = "save-image-canvas";
 const CONTEXT_MENU_SAVE_IMAGE_SVG = "save-image-svg";
 
 const DOWNLOAD_PERMISSION = {
-    permissions: ["downloads"]
+    permissions: ["downloads"],
 };
 
 const qrCode = document.getElementById("qrcode");
 const qrCodePlaceholder = document.getElementById("qrcode-placeholder");
 const qrCodeContainer = document.getElementById("qrcode-container");
-const qrCodeResizeContainer = document.getElementById("qrcode-resize-container");
+const qrCodeResizeContainer = document.getElementById(
+    "qrcode-resize-container"
+);
 const qrCodeText = document.getElementById("qrcodetext");
 
 let resizeMutationObserver;
@@ -49,7 +51,11 @@ let resizeMutationObserver;
 let placeholderShown = true;
 
 // default/last size
-let qrLastSize = 200;
+if (localStorage.getItem("lastSize") === null) {
+    localStorage.setItem("lastSize", 200);
+}
+let qrLastSize = parseInt(localStorage.getItem("lastSize"));
+
 let qrCodeSizeOption = {};
 let savingQrCodeSize = null; // promise
 
@@ -130,7 +136,9 @@ const refreshQrCode = throttle(() => {
  * @returns {bool}
  */
 function isSelected(input) {
-    return input.selectionStart === 0 && input.selectionEnd === input.value.length;
+    return (
+        input.selectionStart === 0 && input.selectionEnd === input.value.length
+    );
 }
 
 /**
@@ -142,7 +150,8 @@ function isSelected(input) {
  * @returns {void}
  */
 function selectAllText(event) {
-    const targetIsSelected = document.activeElement === event.target && isSelected(event.target);
+    const targetIsSelected =
+        document.activeElement === event.target && isSelected(event.target);
 
     console.info("selectAllText", event);
 
@@ -169,7 +178,7 @@ function selectAllText(event) {
  */
 function scrollToTop(event) {
     console.info("scrollToTop", event);
-    event.target.scrollTo(0,0);
+    event.target.scrollTo(0, 0);
 }
 
 /**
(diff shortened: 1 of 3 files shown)


src/popup/qrcode.html
```diff
@@ -22,6 +22,9 @@
 				<img class="icon-dismiss invisible" src="/common/img/close.svg" width="24" height="24" tabindex="0" data-i18n data-i18n-aria-label="__MSG_dismissIconDescription__"></span>
 			</div>
 		</div>
+		<div>
+			<p id="zoomTooltip">Alt +/- to zoom</p>
```
rugk, line 26: That would need to be localized, otherwise, it's static in all languages.

But let's first see whether that is a the appropriate approach.

For more information, please [see the MDN guide on how to localize WebExtensions](https://developer.mozilla.org/en-US/docs/Mozilla/Add-ons/WebExtensions/Internationalization#Providing_localized_strings_in__locales) and [our internationalization (i18n) guide](https://github.com/TinyWebEx/common/blob/master/CONTRIBUTING.md#internationalisation-i18n).

src/popup/modules/UserInterface.js
```diff
@@ -591,72 +629,109 @@ export function init() {
...
+ * Listens to Alt + plus/minus keypresses to alter 
+ * QR code size.
+ *
+ * @function
+ * @returns {void}
+ */
+function zoomQrCode(e) {
+    var evtobj = window.event ? event : e;
+    if (evtobj.keyCode == 61 && evtobj.altKey && qrLastSize < 440) {
+        setNewQrCodeSize(qrLastSize + 30, true);
```
rugk, line 728: Note `setNewQrCodeSize` may – depending on https://github.com/rugk/offline-qr-code/blob/220d6f3c7394ba4dc5a8d831edd3807838041c9a/src/popup/modules/UserInterface.js#L242 save the size to add-on storage's, too. THough thinking about it, that makes sense, because it is then just persisted.

So yeah, I dunno whether my idea of remembering it additionally, was a good idea, but it seems okay'ish.

src/popup/modules/UserInterface.js
```diff
@@ -591,60 +629,109 @@ export function init() {
...
+    return Promise.all([
+        applyingMonospaceFont,
+        applyingQrSize,
+        applyingQrColor,
+        initQrTypespecificSettings,
+    ]);
+}
+
+// zoom listener
+document.onkeydown = zoomQrCode;
```
rugk, line 716: IMHO better use `addEventListener` it's the modern and better syntax.

src/popup/modules/UserInterface.js
```diff
@@ -591,60 +629,109 @@ export function init() {
...
+ * @returns {void}
+ */
+function zoomQrCode(e) {
+    var evtobj = window.event ? event : e;
+    if (evtobj.keyCode == 61 && evtobj.altKey && qrLastSize < 440) {
+        setNewQrCodeSize(qrLastSize + 30, true);
+
+        localStorage.setItem("lastSize", qrLastSize);
+    }
+    if (evtobj.keyCode == 173 && evtobj.altKey && qrLastSize > 50) {
```
rugk, line 732: `keyCode` is kinda deprecated, see the warning at https://developer.mozilla.org/en-US/docs/Web/API/KeyboardEvent/keyCode, so let's better switch to alternatives, if possible.

rugk (changes requested):

rugk (changes requested):
Hi @eridilla,

I am a little critical with having such a non-standard tooltip like <kbd>Alt</kbd>+ (+/-) for zoomimng, given that as I said in https://github.com/rugk/offline-qr-code/issues/189#issuecomment-1284404490 mouse wheel and zoom also already works.

It's a good idea to give a hint, but maybe that is a little too obtrusive (at best it should work intuitively, but I see why that was not possible, as you've explained in https://github.com/rugk/offline-qr-code/issues/189#issuecomment-1356860550).

So the approach I would recommend, as that is not the most important feature here, would be to add a tip via the (already integrated) [RandomTips library](https://github.com/TinyWebEx/RandomTips/).

Also I need to test this out, of course.

---

A general note, please: Next time avoid reformatting many unrelated files, that are not related to the change. The format looks okay, so I will keep that (dunno how that happened? EsLint?), but it makes reviewing harder. Or, at least, do it in a separate commit.

Also, BTW: Next time, try to avoid doing a pull request from the master branch, because you can run into problems when you have a "non-clean" master that does not follow this repo here (i.e. "upstream"). [See this article for details](https://help.github.com/en/articles/proposing-changes-to-your-work-with-pull-requests). Anyway, this is only a tip for the next time. :smiley:

eridilla:
Hey @rugk,
thank you for the review and the tips!

I looked at your comments and fixed accordingly. I rewrote the key press listener to the addEventListener function and replace the deprecated keyCodes. I removed the div with the tooltip and added the tip to Tips.js and the english messages.json.

When I was trying out the random tip I got it to show up incorrectly a few times, then changed my code to look like the other tips and then I couldn't get any tip to show at all so I hope the additions I made are correct.

Thank you once again for your comments and I hope these changes are better :)

src/popup/modules/UserInterface.js
```diff
@@ -712,26 +712,19 @@ export function init() {
...
  * Listens to Alt + plus/minus keypresses to alter 
  * QR code size.
- *
- * @function
- * @returns {void}
  */
-function zoomQrCode(e) {
-    var evtobj = window.event ? event : e;
-    if (evtobj.keyCode == 61 && evtobj.altKey && qrLastSize < 440) {
+document.addEventListener('keydown', function(event) {
```
rugk, line 719: naaaw, I would still keep the function `zoomQrCode` or so you had… You can just reference that here AFAIK.

rugk (approved):
Thanks a lot, yeah that looks good now.

Do you want to add yourself to the Contributors file?

src/popup/modules/UserInterface.js
```diff
@@ -591,60 +629,102 @@ export function init() {
...
+ * QR code size.
+ */
+document.addEventListener('keydown', function(event) {
+    if ((event.key == '=' || event.key == '+') && event.altKey && qrLastSize < 440) {
+        setNewQrCodeSize(qrLastSize + 30, true);
+
+        localStorage.setItem("lastSize", qrLastSize);
+    }
+    if (event.key == '-' && event.altKey && qrLastSize > 50) {
+        setNewQrCodeSize(qrLastSize - 30, true);
```
rugk, line 726: Ah and could you extract the `30` into a `const KEYBOARD_ZOOM_SIZE = 30` constant or so? Magic numbers are not a good practise in source code that should be readable… :upside_down_face:

rugk: Same for `50` which I assume is something as `MINIMAL_ZOOM_SIZE` or so?

Just place the constants at the top, there should be some, already. :wink:

PR #304, opened 2023-03-30
Added feature to copy QR to clipboard

### LINKED ISSUE #285: Save QR SVG to copy-paste clipboard
## Background
I have seen https://github.com/rugk/offline-qr-code/wiki/FAQ#how-to-save-the-qr-code-on-disk
I would love there to be an additional item « Save QR code to clipboard ».

## Proposed solution
Add an item line « Save QR code to clipboard SVG » and « Save QR code to clipboard image ».

## Alternatives
Recognize Ctrl-C, or systematically place the QR-code in the clipboard each time a QR code is being displayed. (In that case, have an add-on setting to select SVG or PNG).

ketelaar:
This closes #285 but it has a caveat. This is because I used the Clipboard API in order to copy the QR code to the clipboard, which does not work by default [in the current Firefox release](https://developer.mozilla.org/en-US/docs/Web/API/ClipboardItem). There is another way to implement this using `document.execCommand('copy')` [like this one](https://stackoverflow.com/a/40547470), but I couldn't get it to work.

When you set `dom.events.asyncClipboard.clipboardItem` to `true` in `about:config` this functions works fine. The issue that is blocking it from [ being activated by default](https://bugzilla.mozilla.org/show_bug.cgi?id=1809106) is [being worked on](https://bugzilla.mozilla.org/show_bug.cgi?id=1712122), so this suggests that the required `about:config` entry might be activated by default some time in the future.

I'm not sure on how to proceed with this, should we maybe add a small description in the options to let users know how to activate it with the current (stable) version of Firefox?

Also, I tested this function on both Windows 11 22H2 and macOS Ventura 13.3 on Firefox 111.0.1 and 112.0b8 (Developer Edition) by pasting the copied image to LibreOffice Writer, which worked fine on both platforms and versions of Firefox.

If something needs to change or is a bit unclear, please let me know, I'll get back to you ASAP.

### DIFF
--- a/src/_locales/en/messages.json
+++ b/src/_locales/en/messages.json
@@ -202,6 +202,15 @@
     "message": "Save QR code as an &SVG…",
     "description": "The context menu entry shown for saving SVG images in the popup with an access key."
   },
+  "contextMenuSaveImageClipboard": {
+    "message": "Save QR code to clipboard…",
+    "description": "The context menu entry shown for saving PNG images (to the clipboard) in the popup"
+  },
+  "contextMenuSaveImageClipboardAccessKey": {
+    "message": "Save QR code to &clipboard…",
+    "description": "The context menu entry shown for saving PNG images (to the clipboard) in the popup with an access key."
+  },
+
 
   // options
   "someSettingsAreManaged": {
--- a/src/_locales/en/messages.json
+++ b/src/_locales/en/messages.json
@@ -362,6 +362,11 @@
     "description": "This is an option shown in the add-on settings."
   },
 
+  "optionOverrideCopy": {
+    "message": "CTRL+C/⌘+C to copy QR code",
+    "description": "This is an option shown in the add-on"
+  },
+
   "optionDebugMode": {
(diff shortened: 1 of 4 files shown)


rugk:
Hi @ketelaar,

Note yu can (automatically) [let issues close](https://docs.github.com/en/issues/tracking-your-work-with-issues/linking-a-pull-request-to-an-issue) when a PR is merged by adding some "magic" text to your PR _body_. (Manually linked it now.)

About the clipboard API problem: did you know that web extensions have their own API for clipboard access? I used that in the emoji picker extension for coping the emoji, IIRC that API was only to be used when the address bar search is enabled and the emoji is intended to be copied there.
https://developer.mozilla.org/en-US/docs/Mozilla/Add-ons/WebExtensions/API/clipboard may help here.

Otherwise https://developer.mozilla.org/en-US/docs/Mozilla/Add-ons/WebExtensions/Interact_with_the_clipboard may help.

rugk:
So I assume you do not want to continue this PR @ketelaar?

PR #305, opened 2023-03-30
Blur QR-code image when asking permission

Yzmoo:


Yzmoo:
I was wondering what the status of this pull request is

src/popup/modules/UserInterface.js
```diff
@@ -378,6 +378,10 @@ export function handleQrError(error) {
 function triggerFileSave(file, filename, requestDownloadPermissions) {
     const downloadPermissionGranted = browser.permissions.contains(DOWNLOAD_PERMISSION);
 
+    // Blur QR code so that it is clear that user did not give permission yet
+    qrCode.style.filter = "blur(5px)"; // add the blur effect
```
rugk, line 382: At least the comment on this line is unnecessary IMHO. Its okay you could leave it, but have a look at the clean code prinicples. One big takeaway is that the code itself should be self-explanatory and structured in way that you actually don't need to comment everything. If the code itself shows what is done there is no advantage in adding a comment that says the same again. Instead such comments can even be bad as they “rot” and may become outdated.

src/popup/modules/UserInterface.js
```diff
@@ -388,6 +392,7 @@ function triggerFileSave(file, filename, requestDownloadPermissions) {
             CommonMessages.showInfo("requestDownloadPermissionForQr");
         }
 
+
```
rugk, line 395:
```suggestion
```

src/popup/modules/UserInterface.js
rugk, line 382: Did you test this actually? If so its a great idea to include a screenshot in the PR body (you can just paste it in there) to demonstrate that and to show how it looks like.

Also we have a quite strict CSP (content security policy, so I question whether adding CSS like that inline should even work). See https://developer.mozilla.org/en-US/docs/Mozilla/Add-ons/WebExtensions/Content_Security_Policy#default_content_security_policy for the default one Mozilla has for extensions and https://extensionworkshop.com/documentation/develop/build-a-secure-extension/ for more context on secure extensions.
Here https://github.com/rugk/offline-qr-code/blob/04a608946ed4a592a3a861157f3ed08c36b6e0ca/src/manifest.json#L43 our current CSP is defined (plus the other manifest files).

The alternative is easy, just using/adding a class. Another advantage is you can name the class semantically good, so that it is clear what it does by its name and so again you have clear code and don't need a comment. 

You cab find a lot of resources about CSS naming: [Guides by example](https://dev.to/ziizium/css-naming-conventions-5gd6) and [mdn with general tips](https://developer.mozilla.org/en-US/docs/Learn/CSS/Building_blocks/Organizing). The important thing is just consistency, however, and that your variable names are explanatory and explain semantics.

src/popup/modules/UserInterface.js
```diff
@@ -411,10 +416,13 @@ function triggerFileSave(file, filename, requestDownloadPermissions) {
             if (usePermissionWorkaround) {
                 // if permission result is there, hide info message
                 CommonMessages.hideInfo();
+
```
rugk, line 419:
```suggestion
```

Also please don't add unnecessary empty lines here

rugk (changes requested):
Hi @serkanalgur ,

And sorry for the delay, I'm currently quite busy with other things so sorry for not handling this earlier.

Note you can (automatically) [let issues close](https://docs.github.com/en/issues/tracking-your-work-with-issues/linking-a-pull-request-to-an-issue) when a PR is merged by adding some "magic" text to your PR _body_. (Manually linked it now.)

PR #312, opened 2023-08-22
[Fix issue] Closes #309 

### LINKED ISSUE #309: The undo button is too far from reset button
Related repository: [TinyWebEx/AutomaticSettings](https://github.com/TinyWebEx/AutomaticSettings/tree/5d11d88a8756c0be70fbf13499d1918f1931a058)

## Bug description
https://github.com/rugk/offline-qr-code/assets/3071003/e09491e2-d549-4b27-82db-3f91313bd7de

Let's say I made a mistake that I accidentally pressed the reset button. I expected some confirmation UI appears, but there was nothing. So I have to undoing options manually from my memory, and while doing it, I find there is actually 'Undo' button at the top of the settings page.

## Steps to reproduce
1. In the settings page, click 'Reset all settings to defaults'
2. No confirmations or undo buttons is shown right at that moment
3. Panic

### Expected behavior
An undo button or a reset confirmation should be in visible area at that moment, right after clicked the reset button.

## Possible solution
I think replacing reset button to 'Undo' would be great, like clicking

```
[[ Reset all ]]
```

is replaced to

```
✔️ All are reset [[ Undo ]]
```
.

Or putting the toast message at the bottom is also okay I think?

renji18:
Fixed issue which Closes #309

### DIFF
--- a/src/options/options.html
+++ b/src/options/options.html
@@ -29,13 +29,7 @@
 				</a>
 				<img class="icon-dismiss invisible" src="/common/img/close.svg" width="24" height="24" tabindex="0" data-i18n data-i18n-aria-label="__MSG_dismissIconDescription__"></span>
 			</div>
-			<div id="messageSuccess" aria-label="success message" data-i18n data-i18n-aria-label="__MSG_ariaMessageSuccess__" class="message-box success invisible fade-hide">
-				<span class="message-text">That worked!</span>
-				<a href="#">
-					<button type="button" class="message-action-button micro-button success invisible"></button>
-				</a>
-				<img class="icon-dismiss invisible" src="/common/img/close.svg" width="24" height="24" tabindex="0" data-i18n data-i18n-aria-label="__MSG_dismissIconDescription__"></span>
-			</div>
+			
 			<div id="messageError" aria-label="error message" data-i18n data-i18n-aria-label="__MSG_ariaMessageError__" class="message-box error invisible fade-hide">
 				<span class="message-text">An error happened.</span>
 				<a href="#">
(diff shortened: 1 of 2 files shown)


renji18:
Hi @rugk, thanks for your help, I was able to successfully link this PR with the issue, thanks again. Please review the code and let me know if this good to be merged or not.

src/options/options.html
```diff
@@ -195,6 +189,13 @@ <h1 data-i18n="__MSG_titleAddonBehavior__">Addon behavior</h1>
 				<hr /><br />
 				<button data-i18n="__MSG_optionsResetButton__" type="button" id="resetButton">Reset all settings to defaults</button>
 			</div>
+			<div id="messageSuccess" aria-label="success message" data-i18n data-i18n-aria-label="__MSG_ariaMessageSuccess__" class="message-box success invisible fade-hide success-undo-message-box">
```
rugk, line 192: Hmm, good idea in theory, but moving all success messages there is likely not a good idea.

The thing is, you actually do not want to show _all_ success messages here, because other modules could show other success messages and they should not be at the bottom. And if you just move the HTML element with id `#messageSuccess` that is just the [the default container for success messages](https://github.com/TinyWebEx/MessageHandler#parameters), but good news: the `MessageHandler` module can have more such arbitrary message locations/HTML elements. 
That is why I've opened https://github.com/TinyWebEx/AutomaticSettings/issues/28 and one would likely have to adjust that as well and then use the library here. (You can do so, it is a git submodule, online guides should tell you how to work with them…) See there for more details.

rugk (changes requested):
So while that solves the problem, we have a one big technical problem with this solution…

src/options/options.html
renji18: So basically, I have to place the HTML I changed to the previous location, create a new id which displays the undo message, and then place a new HTML that only displays the undo message at the bottom. Did I get it right?

rugk: Yeah that's it! The more or less tricky part is to then the fact that likely the AutomaticSettings library (module) needs to be adjusted to work with that message box instead of the success one - of course keep in mind it should at best have no breaking changes for other add-ons using the library, so that should be optional.

https://github.com/TinyWebEx/PermissionRequest does something similar/is used in the same way, as it also can register for arbitrary message boxes.

PR #313, opened 2023-10-17
Update manifest.json

ladiff777:
This comprehensive documentation provides detailed explanations for each section and property in your manifest.json file, making it easier for developers and maintainers to understand and work with the add-on's configuration.

### DIFF
--- a/src/manifest.json
+++ b/src/manifest.json
@@ -1,19 +1,20 @@
 {
-  "manifest_version": 2,
-  "name": "OfflineQr DEV VERSION",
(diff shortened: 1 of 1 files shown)


rugk (commented):
Hi @ladiff777,

However, I have to say, I would not accept this PR like this.
The reason is these comments do not add much value. Clean code etc. practices show you do not need to comment each line, it should be self-explanatory. Most comments here explain not a lot of stuff and even if, some things should just be known to extension developers.

Like, you would not design a clock like this:
ReadibgSourceCodeCommentsUsefullness

Issue #319, opened 2024-05-25
Improve Contribution documentation


elizalflores:
I believe the project `README.md` and `CONTRIBUTING.md`  should include more clarifying information and directions for developers on how to contribute.

Currently there is further, nested documentation on how to contribute to the project located in the TinyWebEx project.

I plan to modify the `README.md` and `CONTRIBUTING.md` to help give more information about these additional guidelines.

rugk:
Sure do a PR; but from your description it is not really clear what you want to modify.

Note the big parts should be into the "nested" TinyWebEx contrib doc, as this applies to all extensions and I don't want to maintain that change in each extension.

PR #320, opened 2024-06-02
[Fix Issue #319] Increase README clarity and references

elizalflores:
Fixes [#319](https://github.com/rugk/offline-qr-code/issues/319)

I've added some lines of text to both the `README.md` and `CONTRIBUTING.md` to give more clarity on how to contribute and provide more context about the project.

Since the common contribution documentation redirects to the TinyWebEx repo, I've added additional context on why interested parties will be redirected there.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -1,7 +1,8 @@
 Nice to see you want to contribute to this project! :+1: :tada:
 Please have a look at this guide to know what you are changing.
 
-### **Note:** This project is an add-on from the TinyWebEx modules. 
+> [!NOTE]
+> This project uses the same guidelines as the TinyWebEx library modules for add-on's. 
 
 There is a master document outlining the common contributing guidelines for this project and other TinyWebEx modules (common add-on files for my add-on(-related) projects). 
 
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -1,9 +1,13 @@
 Nice to see you want to contribute to this project! :+1: :tada:
 Please have a look at this guide to know what you are changing.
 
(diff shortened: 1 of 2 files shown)


CONTRIBUTING.md
```diff
@@ -1,9 +1,13 @@
 Nice to see you want to contribute to this project! :+1: :tada:
 Please have a look at this guide to know what you are changing.
 
-As I do not want to duplicate the instructions all over, please **find the common contributors docs here**: https://github.com/TinyWebEx/common/blob/master/CONTRIBUTING.md
+### **Note:** This project is an add-on from the TinyWebEx modules.
```
rugk, line 4: Not really, I've fixed it. Also I don't really like using headings (#) in markdown for highlighting…

So maybe?

```suggestion
### Note
This project uses the same guidelines as the TinyWebEx library modules for add-on's. 

```

Note nowadays you can also use proper notes etc. in GitHub's markdown, so we probably want to use that?
https://github.com/orgs/community/discussions/16925

(But anyway, let's adjust the text. Offline-QR-Code is not a TinyWebEx thing in itself, it uses TinyWebEx modules, which is an important technical difference, which could otherwise confuse users.)

rugk (changes requested):

CONTRIBUTING.md
elizalflores: Sounds good! I didn't know that feature existed in GitHub's markdown now, thanks for letting me know - I tried implementing it according to the link you provided. Here's the most recent [commit with the requested changes](https://github.com/rugk/offline-qr-code/pull/320/commits/1ab75467bb771cf3cf47b4a0461c6a001698e2f0).

rugk (approved):
Hi @elizalflores,

It now looked fine, so I'll merge this. Thanks a lot, again!

PR #334, opened 2024-10-10
Update messages.json: RO, unkownError field description and translation

ripazha4299:


rugk (commented):
Requesting a potential review from @Criseda. Unless there are objections, this will be merged soon.

Criseda:
That looks good to me 👍 The phrasing is more common on other websites compared to what i put before.

PR #339, opened 2024-10-31
Add new Hacktoberfest features Urdu translations.

Toprun123:
Added the new translations for Urdu from the `en/messages.json` file. The the order of fields couldn't be matched so I left it as is. Could you please accept this before Hacktoberfest ends? @rugk

### DIFF
--- a/src/_locales/ur/messages.json
+++ b/src/_locales/ur/messages.json
@@ -16,6 +16,14 @@
     "message": "افلائن کیو آر کوڈ",
     "description": "The title for the button, which opens the popup."
   },
+  "buttonRequestPermission": {
+    "message": "اجازت دیں",
+    "description": "The button label, used for requesting a permission that is missing."
+  },
+  "permissionRequiredClipboardRead": {
+    "message": "اس خصوصیت کے لیے کلپ بورڈ سے ڈیٹا کاپی کرنے کی اجازت چاہیے۔",
+    "description": "The message shown, when the user wants to use the clipboard to read from"
+  },
   "commandOpenQrPopup": {
     "message": "افلائن کیو آر کوڈ کھلو",
     "description": "Description of the hot key command to open the QR Code popup (by default with Ctrl+Shift+F10)."
@@ -33,6 +41,10 @@
   "couldNotSaveOption": {
     "message": "سٹںگ کو محفوظ نہیں کیا جا سکا۔",
     "description": "When a setting could not be saved."
+  },
+  "couldNotRequestPermission": {
+    "message": "اجازت طلب کرنا ناکام ہوگیا۔",
+    "description": "When the permission request fails."
   },
   "messageUndoButton": {
     "message": "واپس",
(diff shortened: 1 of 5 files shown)


Toprun123 (commented):
Added 8 new translations in urdu.

rugk (approved):
Ah thanks. You also were the initial contributor, so I have nobody else for review, so I'll merge this.

And you get one more PR for Hacktoberfest! :wink:

src/_locales/ur/messages.json
```diff
@@ -170,6 +182,14 @@
   },
 
   // context menu
+  "contextMenuItemConvertPageURL": {
+    "message": "اس صفحے کے لیے کیو آر کوڈ بنائیں",
+    "description": "Context menu item title for generating a QR code for the current page URL."
+  },
+  "contextMenuItemConvertPageURLAccessKey": {
+    "message": "&Generate QR code for this page",
```
rugk, line 190: This is apparently missing a translation though hmm… cannot merge it as that, but I'll add a tag so it counts as accepted :wink:

rugk (changes requested):

rugk:
BTW what do you mean by:
> The the order of fields couldn't be matched so I left it as is. 

Which fields do you mean? If there are placeholders, you can usually just place them where you want (but there should be none in the new strings). Or do you mean the order in the JSON file? That does not really matter, but also looks fine for me, actually.

src/_locales/ur/messages.json
Toprun123: It's because i did not understand what the & symbol is doing.. Should i replace the entire text or leave the &Generate part (there are 5 of these that i missed.)

Toprun123:
> Or do you mean the order in the JSON file? That does not really matter, but also looks fine for me, actually.

That's what I meant, the order in the JSON file.

src/_locales/ur/messages.json
Toprun123: I translated it all to Urdu (Maybe I should have left the word following the `&`?)

rugk: As for the ampersand (`&?) symbol, what this means [is explained here](https://github.com/TinyWebEx/common/blob/master/CONTRIBUTING.md#translations). Basically it's for creating a keyboard hotkey.

rugk (approved):
As for the ampersand symbol see my comment above.
As for our language I am also not sure how these can be translated, but here is what the docs say, which may be relevant:
> In some localized versions of Firefox (Japanese and Chinese), the access key is surrounded by parentheses and appended to the menu label, unless the menu title itself already ends with the access key ("toolkit(&K)" for example). For more details, see [Firefox bug 1647373](https://bugzil.la/1647373).

Actually, you could also [try out the development version](https://github.com/TinyWebEx/common/blob/master/CONTRIBUTING.md#coding) and report back whether it/everything works/you like it or what may still be adjusted, so I can do this before the next release. In case there is something to be done… :smile: 

Just so we know the translation works good in the browser when it is actually used.

Anyway, I'll merge this for now, if you need to adjust/want to adjust something more, please just create a new PR. And thanks again for the translation!

PR #340, opened 2024-10-31
Polish translation updated

roffikk:
Polish translation updated


rugk (approved):
Thanks a lot!

We have no reviewers for this locale except you, so I'll merge it and thanks for your contribution.

BTW note if you want you could also write the Changelog entry already, just copy an existing one [as explained as a side-note here](https://github.com/TinyWebEx/common/blob/master/CONTRIBUTING.md#credit). (Anyway, I'll merge this first.)

PR #342, opened 2024-11-04
add button to copy qrCode text

GuillaumeZaki:
I’ve added a button that allows users to copy the text generated for the QR code. When the text is successfully copied, a message will appear to confirm the action. This is my first contribution

You can find the changes below:

Capture d’écran 2024-11-04 à 12 41 44
Capture d’écran 2024-11-04 à 12 42 30

### DIFF
--- a/src/popup/qrcode.css
+++ b/src/popup/qrcode.css
@@ -125,3 +125,34 @@ html, body {
     color: var(--white-100);
   }
 }
+
+/* Copy button  */
+.copy-container{
+  margin: 5px 0;
+}
+
+.copy-container{
+  display: flex;
+  flex-direction: column;
+  align-items: center;
+}
+
+/* Copy status message  */
+#copy-status{
+  display: none;
+  position: absolute;
+  top: -6px;
+  color: black;
+  background-color: white;
+  padding: 2px;
+  border-radius: 2px;
+  transition: opacity 0.5s ease;
+}
+
+/* copy status message when status is shwon */
+#copy-status.show{
+  display: block;
+  opacity: 1;
+
+}
+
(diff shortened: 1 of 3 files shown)


rugk:
Hi @GuillaumeZaki,

I like how you started contributing and it all seems to work. The question just stands, whether this feature is actually useful/wanted by the users of the extension.

Is there any issue that may be related where users asked for this or so? What is the actual use case behind this? Which problem does it solve?

Because from my experience, I see no reason to copy the _text_ there from that menu. If you select a text and open the QR Code generator from the context menu, you could just directly click "copy". If you use the URL as your example demonstrates you could also just right-click in the browser's address bar and click "copy" there.

We had a similar request in https://github.com/rugk/offline-qr-code/issues/285 for copying the SVG _image_ but that is something totally different. There I can see the usefulness, after all…

I am sorry I cannot give a more positive response, I hope the experience of contributing helped you anyway and you've learned something along the way…

src/popup/qrcode.css
```diff
@@ -125,3 +125,34 @@ html, body {
     color: var(--white-100);
   }
 }
+
+/* Copy button  */
```
rugk, line 129: While this works, it does not look very nice. I would have suggested to style the elements according to (now obsolete but still used here) Firefox Photon style guide like https://web.archive.org/web/20220519062052/https://design.firefox.com/photon/components/message-bars.html

Also maybe a success message using https://github.com/TinyWebEx/MessageHandler (already included here) could have been used instead, which would integrate it more into the existing style.

rugk (changes requested):
I've reviewed it anyway to give you tips for the next time, but I've decided not to merge it. I am sorry, but the idea of the UI/popup interface there is to be as slim as possible in order not to do the one thing it should do: show/generate a QR code. There is no "generate" button by purpose, e.g.

GuillaumeZaki:
Hi Rugk,

Thank you for taking the time to review my pull request and for all the helpful feedback! I appreciate your detailed insights on the feature's utility and the suggestions you gave me for future contributions. This was actually my first contribution to an open-source project, so your advice is especially valuable as I work on improving my skills. 😁

Thanks again for your time and for welcoming new contributors like me. I’m looking forward to contributing more in the future!

PR #348, opened 2025-01-21
Added missing version change notes to the Changelog

niotw:
Added missing version change notes to the Changelog and translated the description property in messages.json for the zh_TW locale.


src/_locales/zh_TW/messages.json
```diff
@@ -2,57 +2,57 @@
   // manifest.json
   "extensionName": {
     "message": "離線 QR 碼產生器",
-    "description": "Name of the extension."
+    "description": "擴展名稱。"
```
rugk, line 5: Please note though the description is not to be translated, aka you don't need to translate it. It's just for further translators.

Maybe to avoid further confusion, better revert this/the changes in this file? I would have nothing against merging them though, if you think it's useful for other translators, but there is really no need to do so!

Issue #308, opened 2023-08-12
Configuration for hiding random tips


xnuk:
First, thanks for making the great extension.

## Background
The tips shows randomly, in [20% chance maybe](https://github.com/TinyWebEx/RandomTips/blob/0eaaf040f21526edb424a40af3fd4c9cbadb3beb/README.md#using)? And it also randomly resizes the popup window, sometimes shows unexpectedly large window like if this message appears:

https://github.com/rugk/offline-qr-code/blob/04a608946ed4a592a3a861157f3ed08c36b6e0ca/src/_locales/en/messages.json#L138

## Proposed solution
Please give a option for hiding the random tips.

## Alternatives
Preserving fixed space for the tips can be an option, instead of randomly hiding tips.

## etc
- Screenshot is hard to take because it... randomly appears.
- (naive idea warning, you can safely ignore this) I think some of tips can be removed, by putting different places. ex)
  - tipSaveQrCode: put a 'Save' button in popup window.
  - tipQrCodeHotkey: move to settings page, and explain how to configure.
  - tipYouLikeAddon, tipTranslateAddon, tipDonate: put as links in either settings page, popup window, or both.

rugk:
Thanks a lot for that detailed (researched) feature request! I see you looked into the source… :wink: 

> tipSaveQrCode: put a 'Save' button in popup window

Good idea, but issue is stuck at finding a good and minimal UI. See/follow https://github.com/rugk/offline-qr-code/issues/225 for details.

> tipQrCodeHotkey: move to settings page, and explain how to configure.

For the given space explaining the latter would be too much, but maybe an entry in the FAQ could help? That _could_ also be linked? (ref https://github.com/rugk/offline-qr-code/issues/67)
For moving well they are shown in both places, actually [it could even be configured how much in both contexts](https://github.com/TinyWebEx/RandomTips/blob/0eaaf040f21526edb424a40af3fd4c9cbadb3beb/examples/Tips.js#L27-L32), but I did not do so, yet. Likely, because, IMHO, it is also relevant to users, who have never seen/used these options before and just may discover an easier way to use the extension.
_Actually_ IMHO that _is_ the main aim of that tip – not to let users change it, that is something not everyone may care about (though may be a good low priority tip) –, but to let users know of that feature initially.

> tipYouLikeAddon, tipTranslateAddon, tipDonate: put as links in either settings page, popup window, or both.

[...]

PR #352, opened 2025-04-09
Fix issue #308: ability for a user to select if he/she wants to see random tips

DovydasGrb1:
According to Issue 308, there was no functionality to disable random tips (which had 20% chance of appearing), therefore, I have added a checkbox in the popup which allows the user to disable random tips.

### DIFF
--- a/src/popup/qrcode.html
+++ b/src/popup/qrcode.html
@@ -47,6 +47,10 @@
 				</div>
 			</div>
 		</div>
+		<label>
+			<input type="checkbox" id="disable-tips-checkbox">
+			Disable random tips
+		  </label>		  
 		<textarea id="qrcodetext"
 					class="browser-style"
 					placeholder="Enter text for QR code here to generate it."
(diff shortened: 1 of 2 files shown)


rugk:
Hi @DovydasGrb1,

Note you can (automatically) [let issues close](https://docs.github.com/en/issues/tracking-your-work-with-issues/linking-a-pull-request-to-an-issue) when a PR is merged by adding some "magic" text to your PR _body_. (Manually linked it now.)

rugk (changes requested):
Please don't add this checkbox to the popup window though. If it is a useful setting for you I'd accept it in the actual **settings** of the extension, but not in the popup. The reason is the popup should be as slim/minimalistic as possible, so a checkbox, which is always shown, is certainly not good. Especially, if, as you correctly explain, the tip is only shown in 20% of the cases (if there is actually any one), but now you have a checkbox "in the way" for 100% of the time. :upside_down_face: 

Additionally, please don't manually call the respective APIs for saving the options, we use a library here for that, but you'll see how that is done in the options HTML etc.. More information/documentation here: https://github.com/TinyWebEx/AutomaticSettings/

PR #353, opened 2025-04-17
Add qr code size warning on too large QR code

### LINKED ISSUE #35: Warning when size is too large for popup window in overflow menu
When the size of the QR code is manually set or remembered, it would be good to show a warning when the maximum size of a popup window there is exceeded.

Otherwise this happens:
grafik

lvchamao:
Pull Request that Closes: #35 

A warning message was added if the fixed qr size is too large to be shown in the maximum dimensions of a firefox popup window (800x600), see screenshot 1. This message does not show on mobile platforms . The warning can be dismissed and is automatically hidden if users manually resize the popup.

Furthermore, when trying large fixed sizes to find overflowing QR Codes, I ran into the issue that when the fixed size is large, e.g. 1000, the initial centering of the QR Code is wrong. (screenshot 2)
I was able to fix this by using the resizeLayout function in the initialization. 

1) 
image

2)
image

### DIFF
--- a/src/_locales/en/messages.json
+++ b/src/_locales/en/messages.json
@@ -108,6 +108,10 @@
     "message": "Use contrasting color",
     "description": "The text of a button that sets a new color with a sufficient contrast."
   },
+  "qrCodeSizeTooLargeWarning": {
+    "message": "The selected QR code size exceeds the available display space. The displayed QR code will be resized to fit. ",
+    "description": "The message shown when the requested QR code size is too large to display and has been automatically scaled down to fit."
+  },
 
   // tips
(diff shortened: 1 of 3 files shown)


rugk (changes requested):
Oh sorry must have overlooked this, but thanks i any case for contributing!

I tried testing this, though, but could not get it to work:

Image

It never showed the warning despite me resizing.

rugk:
Ah okay, now could reproduce. I refactored it a bit to use the `isMobile()` check that uses the native WebExtension API instead of fragile UA parsing, but otherwise looks fine I am merging! Thanks again!
image

